PR #1, opened 2016-10-18
Add support for OSPFv3, BGP on IPv6-only networks & SLAAC

jadinm:


ipmininet/router/config/radvd.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class RADVD(Daemon):
+    """The base class for all Quagga-derived daemons"""
+
+    # Additional parameters to pass when starting the daemon
+    STARTUP_LINE_EXTRA = ''
+    NAME = 'radvd'
+
+    def __init__(self, *args, **kwargs):
```
oliviertilmans: You don't this to be present if all you do is call super

ipmininet/router/config/radvd.py
```diff
@@ -0,0 +1,72 @@
...
+        defaults.debug = ()
+        super(RADVD, self).set_defaults(defaults)
+
+    @property
+    def startup_line(self):
+        s = 'radvd -C {cfg} -p {pid} -m logfile -l {log} -u root {extra}'\
+                .format(cfg=self.cfg_filename,
+                        log=self._file('log'),
+                        pid=self._file('pid'),
+                        extra=self.STARTUP_LINE_EXTRA)
```
oliviertilmans: Same than above, no point in supporting stuff that won't be used.
The extra/STARTUP_LINE_EXTRA should be removed

ipmininet/router/config/radvd.py
```diff
@@ -0,0 +1,72 @@
...
+        return s
+
+    @property
+    def dry_run(self):
+        return 'radvd -c -C {cfg} -u root'\
+               .format(cfg=self.cfg_filename)
+
+    def cleanup(self):
+        process = self._node.popen("killall radvd")  # TODO Find something better
```
oliviertilmans: This should not be needed, as __router.py takes care of starting/stopping the processes in the start/terminate methods.
However since you cannot prevent radvd from forking itself, you indeed need to issue a kill to the child.
Simply read its pid from the pid file you specified (self._file('pid')):
```python
try:
    with open(self._file('pid'), 'r') as f:
        pid = int(f.read())
    self._node.call('kill -9 %d ' % pid)
except (IOError, OSError):
    pass
```

ipmininet/router/config/templates/radvd.mako
```diff
@@ -0,0 +1,33 @@
...
+                    AdvAutonomous on;
+                    AdvValidLifetime ${prefix.valid_lifetime};
+                    AdvPreferredLifetime ${prefix.preferred_lifetime};
+                };
+            % endfor
+                #clients
+                #{
+                #    fe80::21f:16ff:fe06:3aab;
+                #    fe80::21d:72ff:fe96:aaff;
+                #};
```
oliviertilmans: Get rid of the example code

ipmininet/router/config/radvd.py
```diff
@@ -0,0 +1,72 @@
...
+        super(RADVD, self).__init__(*args, **kwargs)
+
+    def build(self):
+        cfg = super(RADVD, self).build()
+        # Update with preset defaults
+        cfg.update(self.options)
+        cfg.debug = self.options.debug
+        # Track interfaces
+        cfg.interfaces = (self._build_interface(itf)
+                          for itf in realIntfList(self._node))
```
oliviertilmans:
```python
for itf in realIntfList(self._node) if itf.ra_prefixes
```

So you will not need to check that property later on.

ipmininet/router/config/templates/radvd.mako
```diff
@@ -0,0 +1,33 @@
+# See `man radvd.conf` for more details
+% for itf in node.radvd.interfaces:
+    % if len(itf.ra_prefixes) != 0:
```
oliviertilmans: You can get rid of that guard, cfr. above).

ipmininet/router/config/radvd.py
```diff
@@ -0,0 +1,72 @@
...
+    @staticmethod
+    def _build_interface(itf):
+        """Build the ConfigDict object representing the interface"""
+        ra_list = []
+        rdnss_list = []
+        for prefix in itf.ra_prefixes:
+            if prefix.get('prefix') is not None:
+                ra_list.append(
+                    ConfigDict(prefix=prefix['prefix'],
+                               valid_lifetime=prefix.get('valid_lifetime',
```
oliviertilmans: There should be some mention of these parameters somewhere in a docstring, i.e. in IPNet.addLink() (for these and all the others you created earlier)

ipmininet/router/config/ospf6.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _build_interfaces(self, interfaces):
+        """Return the list of OSPF6 interface properties from the list of
+        active interfaces"""
+        return [ConfigDict(description=i.describe,
+                           name=i.name,
+                           # Is the interface between two routers?
+                           active=self.is_active_interface(i),
+                           priority=i.get('ospf6_priority',
+                                          self.options.priority),
```
oliviertilmans: I think it would be useful to wrap these and try to get OSPFv2 values if specified, i.e.
```python
priority=i.get('ospf6_priority',
                      i.get('ospf_priority', self.options.priority)),
```

ipmininet/examples/simple_ospfv3_network.py
```diff
@@ -0,0 +1,60 @@
+"""This file contains a simple OSPF topology"""
```
jadinm: Changed

ipmininet/link.py
```diff
@@ -27,6 +27,8 @@ def __init__(self, *args, **kwargs):
         super(IPIntf, self).__init__(*args, **kwargs)
         self.isUp(setUp=True)
         self._refresh_addresses()
+        self.ra_prefixes = kwargs.get('ra', [])
```
jadinm: Changed

ipmininet/router/config/base.py
```diff
@@ -247,11 +247,16 @@ def __init__(self, node, additional_daemons=(), *args, **kwargs):
...
+        from ospf6 import OSPF6
+        # We don't want any zebra-specific settings, so we rely on the OSPF/OSPF6
         # DEPENDS list for that daemon to run it with default settings
         # We also don't want specific settings beside the defaults, so we don't
         # provide an instance but the class instead
-        d = [OSPF]
+        d = []
+        if node.use_v4:
+            d.append(OSPF)
+        elif node.use_v6:
```
jadinm: Changed

ipmininet/router/config/radvd.py
jadinm: Changed

ipmininet/router/config/radvd.py
```diff
@@ -0,0 +1,72 @@
...
+from .utils import ConfigDict
+from ipmininet.utils import realIntfList
+
+RA_DEFAULT_VALID = 86400
+RA_DEFAULT_PREF = 14400
+DEFAULT_ADV_RDNSS_LIFETIME = 25
+
+
+class RADVD(Daemon):
+    """The base class for all Quagga-derived daemons"""
```
jadinm: Changed

ipmininet/router/config/radvd.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Additional parameters to pass when starting the daemon
+    STARTUP_LINE_EXTRA = ''
```
jadinm: Changed

ipmininet/router/config/radvd.py
jadinm: Changed

ipmininet/router/config/radvd.py
jadinm: Changed

ipmininet/router/config/radvd.py
jadinm: Changed

ipmininet/router/config/templates/radvd.mako
jadinm: Changed

ipmininet/router/config/templates/radvd.mako
jadinm: Changed

ipmininet/examples/README.md
```diff
@@ -18,8 +18,7 @@ The following sections will detail the topologies.
 _topo name_ : simple_ospf_network
 _args_ : n/a
 
-This network spawn a single AS topology, using OSPF, with multiple areas and
-variable link metrics.
+This network spawn a single AS topology, using OSPF, with variable link metrics.
```
jadinm: Changed: c77e21136ce17702233f7ced3ccae067dc61b845

ipmininet/router/config/bgp.py
```diff
@@ -170,7 +184,7 @@ def _find_peer_address(base, peer):
             for n in i.broadcast_domain.routers:
                 if n.node.name == peer:
                     ip = n.ip
-                    return (ip if ip else n.ip6), n.node
+                    return (ip, n.node, i.ip > ip) if ip else (n.ip6, n.node, i.ip6 > n.ip6)
```
oliviertilmans: Especially since we now have globally guaranteed router ids

ipmininet/router/config/radvd.py
```diff
@@ -0,0 +1,81 @@
...
+class RADVD(Daemon):
+    """The class representing the radvd daemon, used for router advertisements"""
+
+    NAME = 'radvd'
+
+    def build(self):
+        cfg = super(RADVD, self).build()
+        # Update with preset defaults
+        cfg.update(self.options)
+        cfg.debug = self.options.debug
```
oliviertilmans: I don't see this used anywhere

ipmininet/router/config/ospf.py
```diff
@@ -45,8 +44,9 @@ class OSPF(QuaggaDaemon):
     NAME = 'ospfd'
     DEPENDS = (Zebra,)
 
-    def __init__(self, *args, **kwargs):
-        super(OSPF, self).__init__(*args, **kwargs)
+    def __init__(self, node, routerid=None, *args, **kwargs):
+        super(OSPF, self).__init__(node=node, *args, **kwargs)
+        self.routerid = routerid
```
oliviertilmans: What's that doing there? You don't want routers in an area to have the same router id ...

ipmininet/router/config/templates/bgpd.mako
```diff
@@ -12,17 +12,28 @@ debug bgp section
 router bgp ${node.bgpd.asn}
 % if node.bgpd.routerid:
     bgp router-id ${node.bgpd.routerid}
-% elif node.ospfd:
-    bgp router-id ${node.ospfd.router_id}
+% else:
+    bgp router-id ${node.zebra.routerid}
```
oliviertilmans: This kind of logic should happen in the build() method of the daemons, i.e. check if self.options has a routerid set or fall back to the one assigned by super()

ipmininet/router/config/templates/ospf6d.mako
```diff
@@ -0,0 +1,40 @@
...
+  ipv6 ospf6 instance-id ${intf.instance_id}
+  <%block name="interface"/>
+!
+% endfor
+
+router ospf6
+  % if node.ospf6d.routerid:
+  router-id ${node.ospf6d.routerid}
+  % else:
+  router-id ${node.zebra.routerid}
```
oliviertilmans: Same comment than for bgpd:
This should be handled in build() (with an implicit fallback to super())

ipmininet/router/config/templates/ospfd.mako
```diff
@@ -24,7 +24,11 @@ interface ${intf.name}
 % endfor
 
 router ospf
-  router-id ${node.ospfd.router_id}
+  % if node.ospfd.routerid:
+  router-id ${node.ospfd.routerid}
+  % else:
+  router-id ${node.zebra.routerid}
```
oliviertilmans: See above

ipmininet/router/config/zebra.py
```diff
@@ -63,6 +65,7 @@ def build(self):
         cfg = super(Zebra, self).build()
         # Update with preset defaults
         cfg.update(self.options)
+        cfg.routerid = self.get_routerid()
```
oliviertilmans: I believe router id should be handled by QuaggaDaemon directly, as it is common to every daemon (and that way will be accessible by the subclasses).

ipmininet/router/config/zebra.py
```diff
@@ -83,6 +86,59 @@ def has_started(self):
...
+                    i = to_visit.pop()
+                    if i in visited:
+                        continue
+                    visited.add(i)
+                    for n in i.broadcast_domain.routers:
+                        if self._equal_routerid(n.node):
+                            break  # We need to change the router id
+                        to_visit.extend(realIntfList(n.node))
+                to_visit = realIntfList(self._node) if to_visit else []
+            return self.last_routerid.with_prefixlen.split("/")[0]
```
oliviertilmans: self.last_routerid.ip.compressed will do the same

ipmininet/router/config/zebra.py
```diff
@@ -83,6 +86,59 @@ def has_started(self):
...
+                    and hasattr(d, "routerid") \
+                    and int(d.routerid) == int(self.last_routerid):
+                return True
+        return False
+
+    def get_routerid(self):
+        """Computes the default router id for Quagga daemons
+        It returns the most-visible IPv4 address among its router interfaces.
+        If no IPv4 addresses exists for its router, it generates a unique router id"""
+        ip_list = sorted((ip for itf in realIntfList(self._node)
```
oliviertilmans: The generated router id should be the same for all daemons not manually overwriting it for consistency reasons (i.e. not having to remember a router by 4 different identifiers ...)

ipmininet/router/config/zebra.py
```diff
@@ -83,6 +86,59 @@ def has_started(self):
...
+                          for ip in itf.ips()),
+                         cmp=address_comparator)
+        if len(ip_list) == 0:
```
oliviertilmans: Before looking to assign a new id, you should check that no other routing daemon on this router has one (and copy it if thats the case). Check for options.routerid in all config.daemons instances

ipmininet/router/config/zebra.py
```diff
@@ -83,6 +86,59 @@ def has_started(self):
...
+                         cmp=address_comparator)
+        if len(ip_list) != 0 \
+                and int(ip_list.pop().ip) == int(self.last_routerid):
+            return True
+
+        # Check that a router id explicitly set
+        # in any other Quagga daemon is not in conflict
+        # with the current router id
+        for d in n.config.daemons:
+            if issubclass(d, QuaggaDaemon) and d != self \
```
oliviertilmans: This should not be restricted to QuaggaDaemon only as there are other routing daemons not part of quagga (i.e. if I want to play with babel it won't run as part of quagga).

ipmininet/router/config/zebra.py
```diff
@@ -55,6 +56,7 @@ class Zebra(QuaggaDaemon):
     # We want zebra to preserve existing routes in the kernel RT (e.g. those
     # set via ip route)
     STARTUP_LINE_EXTRA = '-k'
+    last_routerid = ip_interface("0.0.0.1")
```
oliviertilmans: As all you want is an IP address you should be good with ip_address (or IPv4Address) directly

ipmininet/router/config/zebra.py
```diff
@@ -83,6 +86,59 @@ def has_started(self):
         # We override this such that we wait until we have the API socket
         return os.path.exists(self.zebra_socket)
 
+    @classmethod
+    def incr_last_routerid(cls):
```
oliviertilmans: As the routerid should be shared accross daemons on the same router and must be different between router,
there are no benefits into having a per-daemon counter (so static should be fine).

ipmininet/router/config/zebra.py
jadinm: Then, wouldn't it be more sound to push the router id generation to RouterConfig instead of QuaggaDaemon (as suggested in another comment) ?

PR #2, opened 2017-10-29
Add IPv6 support in the CLI

jadinm:
This includes auto-replacement by IPv6 address and ping-related CLI commands that also check IPv6 connectivity.

But this does not include iperf-related CLI commands.

### DIFF
--- a/ipmininet/cli.py
+++ b/ipmininet/cli.py
@@ -31,3 +31,37 @@ class IPCLI(CLI):
             finally:
                 lg.info(n, '|', l)
         lg.info('\n')
+
+    def default(self, line):
+        """Called on an input line when the command prefix is not recognized.
+        Overridden to run shell commands when a node is the first CLI argument.
+        Past the first CLI argument, node names are automatically replaced with
+        corresponding addresses if possible.
+        An IPv4 address will be chosen by default if any and if the network allows IPv4.
+        An IPv6 address will be chosen if any and if the network allows IPv6.
+        In all the other cases, the name is not replaced."""
+
+        first, args, line = self.parseline(line)
+
+        if first in self.mn:
+            if not args:
+                print("*** Enter a command for node: %s <cmd>" % first)
+                return
+            node = self.mn[first]
+            rest = args.split(' ')
+            # Substitute IP addresses for node names in command
+            # If updateIP() returns None, then use node name
+            if self.mn.use_v4:
+                rest = [self.mn[arg].defaultIntf().updateIP() or arg
+                        if arg in self.mn else arg
+                        for arg in rest]
+            if self.mn.use_v6:
+                rest = [self.mn[arg].defaultIntf().updateIP6() or arg
+                        if arg in self.mn else arg
+                        for arg in rest]
+            rest = ' '.join(rest)
+            # Run cmd on node:
+            node.sendCmd(rest)
+            self.waitForNode(node)
+        else:
+            lg.error('*** Unknown command: %s\n' % line)
--- a/ipmininet/cli.py
+++ b/ipmininet/cli.py
@@ -2,6 +2,8 @@
 from mininet.cli import CLI
 from mininet.log import lg
 
+from ipmininet.utils import realIntfList
+
 
 class IPCLI(CLI):
     def do_route(self, line=""):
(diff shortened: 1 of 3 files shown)


ipmininet/ipnet.py
```diff
@@ -350,6 +350,64 @@ def _broadcast_domains(self):
...
+
+           :param hosts: list of hosts or None if all must be pinged
+           :param timeout: time to wait for a response, as string
+           :param returns: the packet loss percentage of IPv4 connectivity if self.use_v4 is set
+                           the loss percentage of IPv6 connectivity otherwise"""
+        ploss_v4 = super(IPNet, self).ping(hosts=hosts, timeout=timeout) if self.use_v4 else 0
+        ploss_v6 = self.ping6(hosts=hosts, timeout=timeout) if self.use_v6 else 0
+
+        return ploss_v4 if self.use_v4 else ploss_v6
+
```
oliviertilmans: This looks overly complex for what it does ...

```python
if self.use_v4:
    return super(IPNet, self).ping(hosts=hosts, timeout=timeout)
else:
    return self.ping6(hosts=hosts, timeout=timeout)
```

ipmininet/ipnet.py
```diff
@@ -350,6 +350,64 @@ def _broadcast_domains(self):
...
+        packets = 0
+        lost = 0
+        if not hosts:
+            hosts = self.hosts
+            log.output('*** Ping6: testing ping6 reachability\n')
+        for node in hosts:
+            log.output('%s -> ' % node.name)
+            for dest in hosts:
+                # We do not try if there is no IPv6 address
+                if node != dest and node.intf().ip6:
```
oliviertilmans: `node.intf().ip6 == node.defaultIntf().ip6`

If that particular interface does not have an IPv6 address, others could ...

ipmininet/ipnet.py
```diff
@@ -350,6 +350,64 @@ def _broadcast_domains(self):
...
+                    opts = ''
+                    if timeout:
+                        opts = '-W %s' % timeout
+                    if dest.intfs and dest.intf().ip6:
```
oliviertilmans: Same comment than above

ipmininet/cli.py
```diff
@@ -31,3 +33,60 @@ def do_ips(self, line):
...
+                    for itf in realIntfList(other_node):
+                        if first_v4 and itf.updateIP() is not None:
+                            ipv4_map[r] = itf.ip
+                        if first_v6:
+                            itf.updateIP6()
+                            for ip6 in itf.ip6s(exclude_lls=True):
+                                ipv6_map[r] = str(ip6.ip)
+                                break
+
+            ip_map = ipv4_map if len(ipv4_map) >= len(ipv6_map) else ipv6_map
```
oliviertilmans: This could be rewritten better to signal an error for dual stacked networks when there is a conflict with the available ip version:

(starting at line 56//Not tested)
```python
def address_pair(n, use_v4=True, use_v6=True):
    v4 = v6 = None
    for itf in realIntfList(n):
        if use_v4 and v4 is None:
            v4 = itf.updateIP()
        if use_v6 and v6 is None:
            itf.updateIP6()
            v6 = next(itf.ip6s(exclude_lls=True), None)
        if (not use_v4 or v4 is not None) and (not use_v6 or v6 is not None):
            break
    return v4, v6
 
            
hops = [h for h in rests if h in self.mn]
v4_support, v6_support = address_pair(first)
v4_map = {}
v6_map = {}
for hop in hops:
    v4_map[hop], v6_map[hop] = address_pair(hop, v4_support is not None, v6_support is not None)
ip_map = v4_map if len(v4_map) >= len(v6_map) else v6_map
if len(ip_map) < len(hops):
    missing = filter(lambda h: h not in ip_map, hops)
    version = 'IPv4' if v4_support else 'IPv6'
    lg.error('*** Nodes', missing, 'have no', version, 'address! Cannot execute the command.\n')
    return
node.sendCmd(' '.join(ip_map[h] for h in hops))
self.waitForNode(node)
```

oliviertilmans:
I'd tweak the ping command to not depend at all on the one from the super class. Instead, similarly to the default command: I'd collect the available addresses from all hosts, and execute the v4 ping or the v6 on a per host basis. 
From there, your actual (currently) ping6 command would dynamically choose between running ping6 or ping for each pair depending on the availability of the addresses; i.e. both ping() and ping6() would use an internal function that would take a input a dist hosts:={hostname: (default_v4, default_v6)}.

PR #14, opened 2019-03-20
Add tests

jadinm:
Here are some tests for the daemons

### DIFF
--- a/ipmininet/tests/test_gre.py
+++ b/ipmininet/tests/test_gre.py
@@ -1,0 +1,29 @@
+"""This module tests GRE tunnels"""
+import time
+
+from ipmininet.clean import cleanup
+from ipmininet.examples.gre import GRETopo
+from ipmininet.ipnet import IPNet
+from . import require_root
+
+
+@require_root
+def test_gre_example():
+    try:
+        net = IPNet(topo=GRETopo())
+        net.start()
+
+        t = 0
+        cmd = "ping -W 1 -c 1 -I 10.0.1.1 10.0.1.2".split(" ")
+        while t < 60 and net["h1"].popen(cmd).wait() != 0:
+            t += 1
+            time.sleep(.5)
+        p = net["h1"].popen(cmd)
+        code = p.wait()
+        out, err = p.communicate()
+        assert code == 0, "Cannot use GRE tunnel.\n" \
+                          "The command '%s' printed: [stdout]\n%s\n[stderr]\n%s" % (cmd, out, err)
+
+        net.stop()
+    finally:
+        cleanup()
--- a/ipmininet/tests/test_iptables.py
+++ b/ipmininet/tests/test_iptables.py
@@ -1,0 +1,38 @@
+"""This module tests iptables"""
+
+from ipmininet.clean import cleanup
+from ipmininet.examples.iptables import IPTablesTopo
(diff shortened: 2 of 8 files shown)


ipmininet/tests/test_gre.py
```diff
@@ -0,0 +1,26 @@
...
+        net = IPNet(topo=GRETopo())
+        net.start()
+
+        t = 0
+        cmd = "ping -W 1 -c 1 -I 10.0.1.1 10.0.1.2".split(" ")
+        while t < 60 and net["h1"].popen(cmd).wait() != 0:
+            t += 1
+            time.sleep(.5)
+        p = net["h1"].popen(cmd)
+        assert p.wait() == 0, "Cannot use GRE tunnel"
```
oliviertilmans: I'd at least attempt there to read the output of that command to say why its not working (or make sure its not working)

ipmininet/tests/test_iptables.py
```diff
@@ -0,0 +1,26 @@
...
+        ip = net["r2"].intf("r2-eth0").ip
+        cmd = "ping -W 1 -c 1 %s" % ip
+        p = net["r1"].popen(cmd.split(" "))
+        assert p.wait() == 0, "pings over IPv4 should not be blocked"
+
+        ip6 = net["r2"].intf("r2-eth0").ip6
+        cmd = "ping6 -W 1 -c 1 %s" % ip6
+        p = net["r1"].popen(cmd.split(" "))
+        assert p.wait() != 0, "pings over IPv6 should be blocked"
+
```
oliviertilmans: Could reverse tests be also added, i.e., to show that _some_ connectivity works (and raise errors if it does not)

ipmininet/tests/test_ospf.py
```diff
@@ -0,0 +1,99 @@
...
+    ({},
+     {"params1": {"ospf_priority": 1}},
+     ["  ip ospf priority 1"]),
+    ({},
+     {"params1": {"ospf_dead_int": "minimal hello-multiplier 2"}},
+     ["  ip ospf dead-interval minimal hello-multiplier 2"]),
+    ({"redistribute": [OSPFRedistributedRoute("connected", 1, 15), OSPFRedistributedRoute("static", 2, 50)]},
+     {},
+     ["  redistribute connected metric-type 1 metric 15",
+      "  redistribute static metric-type 2 metric 50"]),
```
oliviertilmans: This is fine from a config generation point of view, but I'd add a parameter to those tests to describe the expected resulting paths.

ipmininet/tests/test_sshd.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@require_root
+def test_sshd_example():
+    try:
+        net = IPNet(topo=SSHTopo())
+        net.start()
+
+        ssh_key = None
+        with open("/tmp/sshd_r2.cfg") as fileobj:
```
oliviertilmans: We can definitely improve the whole ergonomics of the SSH key handling here. Likely though an helper in the config object.

jadinm: I could just import KEYFILE in ipmininet.router.config.sshd. Isn't it sufficient?
This code is needed because I want to check the content written in the configuration.

oliviertilmans:
LGTM
Thanks!

PR #16, opened 2019-03-29
Python3 compatibility

jadinm:
Mininet has python3 compatibility in its master branch for a few months.
This extends this compatibility to IPMininet.

### DIFF
--- a/ipmininet/iptopo.py
+++ b/ipmininet/iptopo.py
@@ -84,7 +84,7 @@ class IPTopo(Topo):
 
     def routers(self, sort=True):
         """Return a list of router node names"""
-        return filter(self.isRouter, self.nodes(sort))
+        return [n for n in self.nodes(sort) if self.isRouter(n)]
 
     def addOverlay(self, overlay):
         """Add a new overlay on this topology"""
--- a/ipmininet/iptopo.py
+++ b/ipmininet/iptopo.py
@@ -118,7 +118,7 @@ class RouterDescription(str):
 
     def __init__(self, o, topo):
         self.topo = topo
-        super(RouterDescription, self).__init__(o)
+        super(RouterDescription, self).__init__()
 
     def addDaemon(self, daemon, default_cfg_class=BasicRouterConfig,
                   cfg_daemon_list="daemons", **daemon_params):
--- a/ipmininet/iptopo.py
+++ b/ipmininet/iptopo.py
@@ -1,4 +1,6 @@
 """This module defines topology class that supports adding L3 routers"""
+from builtins import str
+
 from mininet.topo import Topo
 from mininet.log import lg
 
--- a/ipmininet/link.py
+++ b/ipmininet/link.py
@@ -1,6 +1,9 @@
 """Classes for interfaces and links that are IP-agnostic. This basically
 enhance the Intf class from Mininet, and then define sane defaults for the link
 classes and a new TCIntf base."""
+from builtins import str
+from ipmininet import basestring
+
 from itertools import chain
 import subprocess
 from ipaddress import ip_interface, IPv4Interface, IPv6Interface
@@ -151,7 +154,7 @@ class IPIntf(_m.Intf):
                     addr = ip_interface(u'%s/%s' % (addr, prefixLen))
                 else:
                     # no prefixLen defaults to full /128 or /32
-                    addr = ip_interface(unicode(addr))
+                    addr = ip_interface(str(addr))
 
             # Prepare assignment commands
             cmds.append('ip address add dev %s %s'
@@ -211,7 +214,7 @@ def _addresses_of(devname, node=None):
     try:
         addrstr = node.cmd(*cmdline)
     except AttributeError:
-        addrstr = subprocess.check_output(cmdline)
+        addrstr = subprocess.check_output(cmdline).decode("utf-8")
     except (OSError, subprocess.CalledProcessError):
         addrstr = None
     if not addrstr:
@@ -240,9 +243,9 @@ def _parse_addresses(out):
         try:
             t = parts[0]
             if t == 'inet':
-                v4.append(IPv4Interface(unicode(parts[1])))
+                v4.append(IPv4Interface(str(parts[1])))
             elif t == 'inet6':
-                v6.append(IPv6Interface(unicode(parts[1])))
+                v6.append(IPv6Interface(str(parts[1])))
             elif 'link/' in t:
                 mac = parts[1]
         except IndexError:
@@ -319,7 +322,7 @@ class PhysicalInterface(IPIntf):
         if node.inNamespace:
             # cfr man ip-link; some devices cannot change of net ns
             if 'netns-local: on' in subprocess.check_output(
-                    ('ethtool', '-k', name)):
+                    ('ethtool', '-k', name)).decode("utf-8"):
                 log.error('Cannot move interface', name, 'into another network'
                           ' namespace!')
         super(PhysicalInterface, self).__init__(name, *args, **kw)
@@ -355,9 +358,9 @@ class GRETunnel(object):
                               tunnel is not established, the kernel will drop
                               by defualt the encapsualted packets."""
         self.if1, self.if2 = if1, if2
-        self.ip1, self.gre1 = (ip_interface(unicode(if1address)),
+        self.ip1, self.gre1 = (ip_interface(str(if1address)),
                                self._gre_name(if1))
-        self.ip2, self.gre2 = (ip_interface(unicode(if2address)),
+        self.ip2, self.gre2 = (ip_interface(str(if2address)),
                                self._gre_name(if2))
         self.bidirectional = bidirectional
         self.setup_tunnel()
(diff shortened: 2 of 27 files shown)


ipmininet/iptopo.py
```diff
@@ -116,7 +118,7 @@ def __new__(cls, value, *args, **kwargs):
 
     def __init__(self, o, topo):
         self.topo = topo
-        super(RouterDescription, self).__init__(o)
+        super(RouterDescription, self).__init__()
```
oliviertilmans: Why did the init() parameter 'o' disappear here?

ipmininet/router/config/base.py
```diff
@@ -2,6 +2,10 @@
 that is able to provide configurations for a set of routing daemons.
 It also defines the base class for a routing daemon, as well as a minimalistic
 configuration for a router."""
+from builtins import map
+from future.utils import iteritems
+from six import string_types
```
oliviertilmans: Not sure if pulling in six is really needed when all we want could be adding in the baselevel __init__.py=
```python
try:
  basestring
except NameError:
  basestring = str
```

ipmininet/link.py
```diff
@@ -261,6 +264,29 @@ def __init__(self, node1, node2, intf=IPIntf, *args, **kwargs):
 TCIntf.__bases__ = (IPIntf,)
 
 
+class OrderedAddress(object):
```
oliviertilmans: You may want to look into [functools.total_ordering](https://docs.python.org/3/library/functools.html#functools.total_ordering) to dreduce the boilerplate there.

TLDR:
```python
@functools.total_orderding
class OrderedAddress(object):
    def __init__(self, addr):
        self.addr = addr

    def __lt__(self, other):
        return address_comparator(self.addr, other.addr) < 0

    def __eq__(self, other):
        return address_comparator(self.addr, other.addr) == 0
```

oliviertilmans:
> Mininet has python3 compatibility in its master branch for a few months.
> This extends this compatibility to IPMininet.

It was always disappointing to see mininet lagging behind in py2 world, glad you could do this.

Did you try to run `2to3` on the codebase to check you did not miss 'obvious' uncompatibilities?

jadinm:
> The changes are surprisingly few and look harmless.

We could reduce them even further by removing all the "from builtins import map, filter".
It would be slightly ineffective for py2 because list(map()) or list(fitler()) copy the list
but I do not think we will ever exceed the 100 items in these occurrences.
What do you think?

> Did you try to run `2to3` on the codebase to check you did not miss 'obvious' uncompatibilities?

I just checked and I forgot the metaclasses ^^
So, I added a new commit to the Pull Request.
The rest was already made.

ipmininet/iptopo.py
jadinm: Because Python3 raises the following exception if the argument is present:
```
TypeError: object.__init__() takes no parameters
```

oliviertilmans: What are we missing as a functionnality due to that then?

jadinm: None, I think. Python2.7 was already complaining about that in a warning. Python3 enforces it.
This creates a perfectly fine str object.

oliviertilmans:
Then what was it doing there in the first place?
Should we store o as self.o and then override __str__ and __repr__ to return self.o (i.e. not subclass str anymore?)
________________________________

jadinm:
> Then what was it doing there in the first place?

Passing ``o`` to the ``super().__init__()`` was pointless from the very beginning.
This method never took any argument.
But I had not noticed the warning issued by Python2 in the ocean of logs of Mininet.
I noticed it while coding this PR.

> Should we store o as self.o and then override __str__ and __repr__ to return self.o (i.e. not subclass str anymore?)

We don't need to do that.
Removing the parameter does not prevent the RouterDescription object of being used as a str object.
The following code
```
r1 = self.addRouter('r1')
print("%s - %s - %s" % (r1.count("r"), r1.count("a"), r1.upper()))
```
will print
```
1 - 0 - R1
```

Moreover the other tests are working fine as well.

jadinm:
I suggest waiting for PR #18 before merging.
This way I would also check python3 compatibility for the new daemon as well.

jadinm:
I extended the PR for OpenR and the installation scripts.

ipmininet/router/config/base.py
```diff
@@ -35,7 +39,7 @@ def __init__(self, node, daemons=(), sysctl=None,
                        interfaces."""
         self._node = node  # The node for which we will build the configuration
         self._daemons = {}  # Active daemons
-        map(self.register_daemon, daemons)
+        list(map(self.register_daemon, daemons))
```
oliviertilmans: The map call there was a (pointless) way to make this a one-liner instead of a simple for loop.
You definitely do not need the list expansion (but still need to run the loop).

I'd rewrite all such case by their for loop equivalent ...

```python
for d in daemons:
    self.register_daemon(d)
```

ipmininet/router/config/base.py
```diff
@@ -51,23 +55,23 @@ def build(self):
         self._cfg.password = self._node.password
         self._cfg.name = self._node.name
         # Check that all daemons have their dependencies satisfied
-        map(self.register_daemon,
-            (c for cls in self._daemons.values()
-             for c in cls.DEPENDS if c.NAME not in self._daemons))
+        list(map(self.register_daemon,
+                 (c for cls in list(self._daemons.values())
+                  for c in cls.DEPENDS if c.NAME not in self._daemons)))
```
oliviertilmans: Same than above (unneeded map in the first place)

ipmininet/router/config/bgp.py
```diff
@@ -54,7 +56,7 @@ def bgp_fullmesh(topo, routers):
     :param routers: The set of routers peering within each other"""
     def _set_peering(x):
         bgp_peering(topo, x[0], x[1])
-    map(_set_peering, itertools.combinations(routers, 2))
+    list(map(_set_peering, itertools.combinations(routers, 2)))
```
oliviertilmans: unneeded map

ipmininet/router/config/bgp.py
```diff
@@ -118,7 +120,7 @@ class AddressFamily(object):
     def __init__(self, af_name, redistribute=(), networks=(),
                  *args, **kwargs):
         self.name = af_name
-        self.networks = map(ip_network, map(unicode, networks))
+        self.networks = list(map(ip_network, map(str, networks)))
```
oliviertilmans: I'd like to avoid the double list expansion that happens in py2 were possible.
List comprehensions would fit the bill here:

```python
self.networks = [ip_network(str(n)) for n in networks]
```

ipmininet/link.py
```diff
@@ -150,7 +154,7 @@ def _set_ip(self, ip, prefixLen=None):
                     addr = ip_interface(u'%s/%s' % (addr, prefixLen))
                 else:
                     # no prefixLen defaults to full /128 or /32
-                    addr = ip_interface(unicode(addr))
+                    addr = ip_interface(str(addr))
```
oliviertilmans: Did you test this in py2?

IIRC the [python2 ipaddress module backport](https://pypi.org/project/ipaddress/) will straight out [refuse non-unicode input strings](https://github.com/phihag/ipaddress/blob/master/README.md).

ipmininet/router/config/iptables.py
```diff
@@ -38,7 +40,7 @@ def set_defaults(self, defaults):
     def build(self):
         cfg = super(IPTables, self).build()
         table_name = attrgetter('table')
-        cfg.rules = {k: map(str, v)
+        cfg.rules = {k: list(map(str, v))
```
oliviertilmans: Similar than below, rewrite this as a list comprehension instead

ipmininet/router/config/ospf.py
```diff
@@ -30,7 +32,7 @@ def area(self, x):
     def apply(self, topo):
         # Add all links for the routers
         for r in self.nodes:
-            self.add_link(*map(lambda x: (r, x), topo.g[r].keys()))
+            self.add_link(*list(map(lambda x: (r, x), topo.g[r].keys())))
```
oliviertilmans:
```python
self.add_link(*((r, x) for x in top.g[r].keys()))
```

ipmininet/link.py
jadinm: All the code was tested on Python 2.7, 3.5, 3.6 and 3.7.
This line is not covered by tests, but you have a similar modification in ipmininet/utils.py, line 77 which is covered by the tests.

The "str" is not the python2 "str". It is imported from "builtins" (which is installed with the future package) at the top of the file.
This function consistently returns a unicode string no matter the python version.

If you still need to be conviced, this code works in python2.7 and the latest version of ipaddress.
```python
import ipaddress
from builtins import str
ipaddress.ip_address(str(b"1.1.1.1"))
```
while the following doesn't
```
import ipaddress
ipaddress.ip_address(str(b"1.1.1.1"))
˘˘˘

oliviertilmans: I did not notice the str override through import.
What a nasty/clever monkey patch.

PR #20, opened 2019-04-11
Add documentation for pytest

butjar:
Just a quick documentation for how to run the testsuite..

PR #21, opened 2019-04-11
[test] Add documentation for the testsuite

butjar:
Encourage people to run tests before adding code.

### DIFF
--- a/ipmininet/tests/README.md
+++ b/ipmininet/tests/README.md
@@ -1,0 +1,20 @@
+# Ipmininet tests
+The [pytest](https://docs.pytest.org/en/latest/index.html) framework is used
+for the testsuite and are [integrated within
+setuptools](https://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner).
+Currently the suite has end-to-end tests that check if the daemons work as
+expected. Therefore, the tests require an operating environment, i.e.  daemons
+have to be installed and must be in PATH. 
+
+## Run testsuite
+To run the whole testsuite go the top level directory and run:
+
+```
+sudo python setup.py test
+```
+
+You can also run a single test by passing options to pytest:
+
+```
+python setup.py test --addopts "ipmininet/tests/test_sshd.py --fulltrace"
+```


butjar:
Fixed and rebased #20

oliviertilmans:
Thanks!

Before merging, I'd like to get input from @jadinm whether we should directly link on the main README towards a VM that would contain *all* daemons (i.e., maintain a setup script that would install everything). This should then be referenced here as well.
On the plus side, this would let us add some CI ...

butjar:
That would be great. Maybe something like this:
https://github.com/mininet/mininet/blob/master/util/install.sh

Building on top of the [mininet VM](http://mininet.org/download/#option-1-mininet-vm-installation-easy-recommended) would be reasonable, I guess. There are also [scripts for building the VM](https://github.com/mininet/mininet/tree/master/util/vm).

jadinm:
I agree with you: having a VM with everything packaged would be great.
However, I am not sure yet where we could store the VM.
But, sure, we could maintain a script to install the software and ask users to run it on the Mininet VM.

butjar:
> I agree with you: having a VM with everything packaged would be great.
> However, I am not sure yet where we could store the VM.
> But, sure, we could maintain a script to install the software and ask users to run it on the Mininet VM.

Maybe github releases is enough: https://help.github.com/en/articles/distributing-large-binaries

butjar:
> I agree with you: having a VM with everything packaged would be great.
> However, I am not sure yet where we could store the VM.
> But, sure, we could maintain a script to install the software and ask users to run it on the Mininet VM.

Maybe github releases is enough: https://help.github.com/en/articles/distributing-large-binaries

butjar:
> I agree with you: having a VM with everything packaged would be great.
> However, I am not sure yet where we could store the VM.
> But, sure, we could maintain a script to install the software and ask users to run it on the Mininet VM.

Maybe github releases is enough: https://help.github.com/en/articles/distributing-large-binaries

butjar:
Another option is creating a vagrant box and upload it to [Vagrant Cloud](https://app.vagrantup.com/boxes/search)

jadinm:
Using the release option will be a bit risky, my VM .ova is 1.9GB while the limit is 2GB for a given file inside a release. If it increases, we couldn't release it.

But using Vagrant seems cool :)
I built a [vagrant box](https://app.vagrantup.com/ipmininet/boxes/ubuntu-16.04) that you can reference here. If both Vagrant and Virtualbox are installed, they can use ipmininet with:

```
vagrant init ipmininet/ubuntu-16.04
vagrant up
vagrant ssh
```

butjar:
> Using the release option will be a bit risky, my VM .ova is 1.9GB while the limit is 2GB for a given file inside a release. If it increases, we couldn't release it.

Yes ... I would suggest to build an ISO image instead of the OVA. Ubuntu server which is rather huge has less than 1Gig (even with additional stuff, I guess). But vagrant should be enough. Would be cool to have the Vagrantbox build integrated in travis-ci ... I will have a look how to do it as I find time. 

> But using Vagrant seems cool :)
> I built a [vagrant box](https://app.vagrantup.com/ipmininet/boxes/ubuntu-16.04) that you can reference here. If both Vagrant and Virtualbox are installed, they can use ipmininet with:
> 
> ```
> vagrant init ipmininet/ubuntu-16.04
> vagrant up
> vagrant ssh
> ```

It seems like the box has not been released yet:

```
This box has no released versions. It will not be available from vagrant box add, nor will it show up in search results.
```
As soon as we have #25  upstream we can update this documentation.

jadinm:
> It seems like the box has not been released yet

Oops... It is fixed

ipmininet/tests/README.md
```diff
@@ -0,0 +1,20 @@
...
+setuptools](https://docs.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner).
+Currently the suite has end-to-end tests that check if the daemons work as
+expected. Therefore, the tests require an operating environment, i.e.  daemons
+have to be installed and must be in PATH. 
+
+## Run testsuite
+To run the whole testsuite go the top level directory and run:
+
+```
+sudo python setup.py test
```
jadinm: It seems that this command has some issue with the default version of setuptools in Ubuntu.
This can be solved by upgrading it but to avoid complications, I would instead advise users to do:
```
sudo pytest
```

ipmininet/tests/README.md
```diff
@@ -0,0 +1,20 @@
...
+```
+
+You can also run a single test by passing options to pytest:
+
+```
+python setup.py test --addopts "ipmininet/tests/test_sshd.py --fulltrace"
```
jadinm: Same as above:
```
sudo pytest ipmininet/tests/test_sshd.py --fulltrace
```

jadinm:
This commit is merged in PR #33

PR #24, opened 2019-04-13
[WIP] Create installation script for vm environment

butjar:
This PR adds a install script to provision a fresh [ubuntu 16.04 server VM](http://releases.ubuntu.com/xenial/ubuntu-16.04.6-server-amd64.iso) for ipmininet.

The script is the first step towards a CI integrated VM creation and testing e.g. for each release.

PR #25, opened 2019-04-15
Installation script

jadinm:
This adds an installation script in Python.
It will be more readable than a bash script if it grows in the future and it is easier to support multiple discributions.
For the moment, Ubuntu, Debian and Fedora work with this script.

There is also a provisioning script for an Ubuntu 16.04 VM adapted from PR #24 to use the installation script.

A [vagrant box](https://app.vagrantup.com/ipmininet/boxes/ubuntu-16.04) was generated.

### DIFF
--- a/util/build_vm.sh
+++ b/util/build_vm.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+# This script is intended to install Mininet and IPMininet into
+# a brand-new Ubuntu virtual machine,
+# to create a fully usable "tutorial" VM.
+set -e
+
+export LC_ALL=C
+
+MN_VERSION="master"
+MN_INSTALL_SCRIPT_REMOTE="https://raw.githubusercontent.com/mininet/mininet/${MN_VERSION}/util/vm/install-mininet-vm.sh"
+DEPS="python \
+      python-pip \
+      python3 \
+      python3-pip
+      git"
+
+# Upgrade system and install dependencies
+sudo apt update -yq && sudo apt upgrade -yq
+sudo apt install -yq ${DEPS}
+
+# Set mininet-vm in hosts since mininet install will change the hostname
+sudo sed -i -e 's/^\(127\.0\.1\.1\).*/\1\tmininet-vm/' /etc/hosts
+
+# Install mininet
+pushd $HOME
+source <(curl -sL ${MN_INSTALL_SCRIPT_REMOTE})
+sudo pip2 install mininet/
+
+# Install ipmininet
+git clone https://github.com/cnp3/ipmininet.git
+pushd ipmininet
+sudo python util/install.py -ia
+popd
+popd
+
+# Fix setuptools version issue
+sudo pip2 install --upgrade pip
+sudo apt-get remove -y python-pip
+sudo pip2 install --upgrade setuptools
--- a/util/install.py
+++ b/util/install.py
@@ -1,0 +1,132 @@
+import argparse
+import os
+import sys
+
+from utils import supported_distributions, identify_distribution, sh
+
+MininetVersion = "master"
+QuaggaVersion = "1.2.4"
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Install IPMininet with its dependencies")
+    parser.add_argument("-o", "--output-dir", help="Path to the directory that will store the dependencies",
+                        default=os.environ["HOME"])
+    parser.add_argument("-i", "--install-ipmininet", help="Install IPMininet", action="store_true")
+    parser.add_argument("-m", "--install-mininet", help="Install the last version of mininet and its dependencies",
+                        action="store_true")
+    parser.add_argument("-a", "--all", help="Install all daemons", action="store_true")
+    parser.add_argument("-q", "--install-quagga", help="Install Quagga (version %s) daemons" % QuaggaVersion,
+                        action="store_true")
+    parser.add_argument("-r", "--install-radvd", help="Install the RADVD daemon",
+                        action="store_true")
+    parser.add_argument("-s", "--install-sshd", help="Install the OpenSSH server", action="store_true")
+    return parser.parse_args()
+
+
+def install_mininet():
+    dist.install("git")
+
+    if dist.NAME == "Fedora":
+        mininet_opts = "-fnp"
+        dist.install("openvswitch", "openvswitch-devel", "openvswitch-test")
+        sh("systemctl enable openvswitch")
+        sh("systemctl start openvswitch")
+    else:
+        mininet_opts = "-a"
+
+    sh("git clone https://github.com/mininet/mininet.git",
+       "git checkout %s" % MininetVersion,
+       "mininet/util/install.sh %s -s ." % mininet_opts,
+       "pip2 -q install mininet/",
+       cwd=args.output_dir)
+
+
+def install_quagga():
+    dist.install("autoconf", "automake", "libtool", "make", "gcc", "gawk", "pkg-config")
+
+    if dist.NAME == "Ubuntu" or dist.NAME == "Debian":
+        dist.install("libreadline-dev", "libc-ares-dev")
+    elif dist.NAME == "Fedora":
+        dist.install("readline-devel", "c-ares-devel")
+
+    quagga_src = os.path.join(args.output_dir, "quagga-%s" % QuaggaVersion)
+    quagga_tar = quagga_src + ".tar.gz"
+    sh("wget http://download.savannah.gnu.org/releases/quagga/quagga-%s.tar.gz" % QuaggaVersion,
+       "tar -zxvf '%s'" % quagga_tar,
+       cwd=args.output_dir)
+
+    quagga_install = os.path.join(args.output_dir, "quagga")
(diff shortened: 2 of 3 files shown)


util/build_vm.sh
```diff
@@ -0,0 +1,34 @@
...
+sudo apt update -yq && sudo apt upgrade -yq
+sudo apt install -yq ${DEPS}
+
+# Set mininet-vm in hosts since mininet install will change the hostname
+sudo sed -i -e 's/^\(127\.0\.1\.1\).*/\1\tmininet-vm/' /etc/hosts
+
+# Install mininet
+pushd $HOME
+source <(curl -sL ${MN_INSTALL_SCRIPT_REMOTE})
+sudo pip2 install mininet/
```
butjar: I think the `install-mininet-vm.sh` script already does that. Is there any reason for installing with pip? Do you want to ensure python2 is used ?

util/build_vm.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# This script is intended to install Mininet and IPMininet into
+# a brand-new Ubuntu virtual machine,
+# to create a fully usable "tutorial" VM.
+set -e
+
+export LC_ALL=C
+
+MN_INSTALL_SCRIPT_REMOTE="https://raw.githubusercontent.com/mininet/mininet/master/util/vm/install-mininet-vm.sh"
```
butjar: Maybe we should fix a version e.g. `https://github.com/mininet/mininet/blob/2.2.2/util/vm/install-mininet-vm.sh`

butjar:
closes #24

butjar:
> A vagrant box was generated.

:+1:

butjar:
> You could add the [keyword](https://help.github.com/en/articles/closing-issues-using-keywords)
> `Closes #24`
> to the PR description.

Hmm, not sure if it works only with issues ... but you can close #24

util/install.py
```diff
@@ -0,0 +1,130 @@
...
+        dist.install("openvswitch", "openvswitch-devel", "openvswitch-test")
+        sh("systemctl enable openvswitch")
+        sh("systemctl start openvswitch")
+    else:
+        mininet_opts = "-a"
+
+    sh("git clone https://github.com/mininet/mininet.git",
+       "mininet/util/install.sh %s -s ." % mininet_opts,
+       "pip2 -q install mininet/",
+       cwd=args.output_dir)
```
butjar: Maybe pin also the mininet version

util/install.py
```diff
@@ -0,0 +1,130 @@
...
+    quagga_install = os.path.join(args.output_dir, "quagga")
+    sh("./configure '--prefix=%s'" % quagga_install,
+       "make",
+       "make install",
+       cwd=quagga_src)
+
+    sh("rm -r '%s' '%s'" % (quagga_src, quagga_tar))
+
+    sh("groupadd quagga", may_fail=True)
+    sh("usermod -a -G quagga root", may_fail=True)
```
butjar: We could follow mininet and create a dedicated `mininet` user

util/build_vm.sh
```diff
@@ -0,0 +1,34 @@
...
+DEPS="python \
+      python-pip \
+      python3 \
+      python3-pip
+      git"
```
butjar: On my `Ubuntu 16.0.4.6 server` VM running the tests fails with:

```
mininet@mininet:~/ipmininet$ sudo python setup.py test
/usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'python_requires'
  warnings.warn(msg)
zip_safe flag not set; analyzing archive contents...

Installed /home/mininet/ipmininet/.eggs/pytest_runner-4.4-py2.7.egg
running pytest
Traceback (most recent call last):
  File "setup.py", line 39, in <module>
    url='https://github.com/oliviertilmans/ipmininet'
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 970, in run_command
    cmd_obj = self.get_command_obj(command)
  File "/usr/lib/python2.7/distutils/dist.py", line 846, in get_command_obj
    cmd_obj = self.command_obj[command] = klass(self)
  File "/usr/lib/python2.7/dist-packages/setuptools/__init__.py", line 132, in __init__
    _Command.__init__(self, dist)
  File "/usr/lib/python2.7/distutils/cmd.py", line 64, in __init__
    self.initialize_options()
  File "build/bdist.linux-x86_64/egg/ptr.py", line 97, in initialize_options
  File "build/bdist.linux-x86_64/egg/ptr.py", line 108, in ensure_setuptools_version
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 943, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 834, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.VersionConflict: (setuptools 20.7.0 (/usr/lib/python2.7/dist-packages), Requirement.parse('setuptools>=27.3'))
``` 
This is why I didn't installed pip from the system in #24.

butjar:
Besides from the setuptools version problem the scripts work like a charm. Awesome work :+1: .

util/build_vm.sh
jadinm: Pip cannot see mininet if it is not installed through it.
And since mininet is a requirement, pip refuses to install ipmininet.

I could just install ipmininet without pip but a pip installation is cleaner than an easy_install one: https://packaging.python.org/discussions/pip-vs-easy-install/

util/build_vm.sh
jadinm: This is because I would like to add Python 3 compatibility (PR #16) in the near future and only the Mininet master branch is python 3 compatible at the moment.

However, I can add a variable to the script that sets the version of Mininet and let it at "master" for the moment. Then, in the future, fixing to a particular version of Mininet would be easier.

util/install.py
jadinm: I am not convinced of the added value of a new user.
IPMininet scripts will have to be run by the root user, not the ``mininet`` one.

butjar: I agree, just for convenience. However, the mininet vm doesn't have sudo though.

util/build_vm.sh
butjar: :+1:

util/build_vm.sh
butjar: I see, I was using setuptools for both. As long as it is consistent.

util/build_vm.sh
jadinm: I fixed it by adding the following code at the end of the script.

```bash
sudo pip2 install --upgrade pip
sudo apt-get remove -y python-pip
sudo pip2 install --upgrade setuptools
```

util/build_vm.sh
```diff
@@ -0,0 +1,40 @@
...
+git clone https://github.com/cnp3/ipmininet.git
+pushd ipmininet
+sudo python util/install.py -ia
+popd
+popd
+
+# Fix setuptools version issue
+sudo pip2 install --upgrade pip
+sudo apt-get remove -y python-pip
+sudo pip2 install --upgrade setuptools
```
butjar: You can make the script executable

jadinm: Done

util/install.py
oliviertilmans: Quagga recommends to use another user as the daemon will run and expose itself to the network. This is not our primary use-case hence we do not that (+ this would require a bunch of fixes to ensures that all configs are accessible by that new quagga user etc).

Shouldn't we anyway move away from Quagga to FRR at some point?

jadinm: > Shouldn't we anyway move away from Quagga to FRR at some point?

I could change it by FRRouting (I already tested it) but this would first require the modification of the ipmininet BGP template because of this [issue](https://github.com/FRRouting/frr/issues/1644).

util/install.py
```diff
@@ -0,0 +1,132 @@
...
+    dist.install("radvd")
+
+if args.all or args.install_sshd:
+    dist.install("openssh-server")
+
+# Install IPMininet
+
+if args.install_ipmininet:
+    ipmininet_folder = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+    sh("pip2 -q install %s/" % ipmininet_folder)
```
butjar: Don't you need `-e` to install from local source. I have the feeling pip still tries to install from PyPI.

jadinm: That's why I added a slash after ``%s``.
If you have a slash in the name, pip understands that it is a path and not a package name.

If I add ``-e`` it means that the folder will be in developer mode and the code won't be copied to a /\<pip-install-folder>/dist-package/ipmininet.

jadinm: I checked the Vagrant and it seems that it works.
There is a folder ``tests`` while in the pypi version, there is not.

butjar: Thanks for checking

butjar: Confirmed, just didn't reinstall my package with the same version number.

PR #30, opened 2019-04-25
[OpenR] Enhance the install script for OpenR

butjar:
Enhance the install script to build and install OpenR. To run OpenR in
ipmininet "routers" there are two main requirements for the host system:

1. OpenR and its dependencies are installed
2. IPv6 is enabled (disabled by the mininet install script)

The OpenR build requires several steps:

1. Clone OpenR
2. Generate the build script from python scripts
3. Make the generated build script executable
4. Build and install OpenR using the build script

IPv6 support is re-enabled by changing/ removing the kernel parameters
in `/etc/default/grub` and `/etc/sysctl.conf`.

### DIFF
--- a/util/install.py
+++ b/util/install.py
@@ -1,12 +1,16 @@
 import argparse
 import os
 import sys
+import stat
+import re
 
 from utils import supported_distributions, identify_distribution, sh
 
 MininetVersion = "master"
 QuaggaVersion = "1.2.4"
+OpenrRelease = "rc-20190419-11514"
 
+os.environ["PATH"] = "%s:/sbin:/usr/sbin/:/usr/local/sbin" % os.environ["PATH"]
 
 def parse_args():
     parser = argparse.ArgumentParser(description="Install IPMininet with its dependencies")
@@ -21,6 +25,11 @@ def parse_args():
     parser.add_argument("-r", "--install-radvd", help="Install the RADVD daemon",
                         action="store_true")
     parser.add_argument("-s", "--install-sshd", help="Install the OpenSSH server", action="store_true")
+    parser.add_argument("-6", "--enable-ipv6", help="Enable IPv6", action="store_true")
+    parser.add_argument("-f", "--install-openr",
+                        help="Install OpenR. OpenR is not installed with '-a' option since the build takes quite long.\
+                        We also experienced that the build requires a substantial amount of memory (~4GB).",
+                        action="store_true")
     return parser.parse_args()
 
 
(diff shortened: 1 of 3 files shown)


jadinm:
Thanks for the PR!

However, I was unable to execute it correctly on a Ubuntu 16.04 VM.
It seems that 3GB is not sufficient to run.
The building of the OpenR library fails because gcc is killed by my VM due to the lack of memory.

How much RAM did you use to compile OpenR?

butjar:
> Thanks for the PR!
> 
> However, I was unable to execute it correctly on a Ubuntu 16.04 VM.
> It seems that 3GB is not sufficient to run.
> The building of the OpenR library fails because gcc is killed by my VM due to the lack of memory.

I know it's huge unfortunately. I didn't measure, but I think it takes more than an hour with all dependencies on a single core. 

> How much RAM do you have when you compile OpenR?

For me 4 GB work fine.

Do you suggest packaging it though? We can disable it by default (`-a`).

butjar:
Re-enabling IPv6 should be valuable for ipmininet anyway.

jadinm:
> Do you suggest packaging it though? We can disable it by default (``-a``).

I don't think there is an OpenR package and I don't think that we should maintain such package.
So yes, I would remove it from the default build (``-a``).

I would also put in the documentation of the options (in the ``parse_args()`` function) that:
 - OpenR is not included in ``-a``
 - OpenR requires 4GB of RAM to build (because it is a bit frustrating to discover it after an hour ^^)

> Re-enabling IPv6 should be valuable for ipmininet anyway.

Agreed

butjar:
:+1:  I will update the PR

jadinm:
I will test it again later ;)

Also, I have a question. Which distributions are supported by OpenR ?
If I rely on the name of the script ``debian_system_builder.py``, I guess that only Ubuntu and Debian are supported but not Fedora.

If this is the case, you should check the distrib and only install if Ubuntu and Debian with something like:
```python
if dist.NAME != "Ubuntu" and dist.NAME != "Debian":
  print("We only support OpenR installation on Ubuntu and Debian")
else:
  install_openr()
```

butjar:
> I will test it again later ;)
> 
> Also, I have a question. Which distributions are supported by OpenR ?
> Based on the name of the script `debian_system_builder.py`, I guess that only Ubuntu and Debian are supported but not Fedora.

Yes, the build script uses apt to install dependencies. The README says they test on ubuntu16.04. I created the build script generator since the official build toolchain was docker only. :thinking: Maybe I can have a look if replacing the package manager calls is sufficient for a `fedora_system_builder.py`. 

> If this is the case, you should check the distrib and only install if Ubuntu and Debian with something like:
> 
> ```python
> if dist.NAME != "Ubuntu" and dist.NAME != "Debian":
>   print("We only support OpenR installation on Ubuntu and Debian")
> else:
>   install_openr()
> ```

Good point.

util/install.py
```diff
@@ -83,6 +87,49 @@ def install_quagga():
         break
 
 
+def install_openr(openr_release=OpenrRelease,
+                  openr_remote="https://github.com/facebook/openr.git"):
+    openr_install = os.path.join(args.output_dir, "openr")
+    openr_build = os.path.join(openr_install, "build")
+    openr_buildscript = os.path.join(openr_build, "build_openr_debian.sh")
+    debian_system_builder = "debian_system_builder/debian_system_builder.py"
+    sh("git clone %s" % openr_remote,
```
jadinm: It would be nice to install git before cloning of the openr repository.

```suggestion
    dist.install("git")
    sh("git clone %s" % openr_remote,
```

jadinm:
I tested it. It works on Ubuntu 16.04.
However, the script does not work in Debian Stretch.
The installation script tries to install ``gcc-5`` which does not exist in this distribution.

butjar:
> I tested it. It works on Ubuntu 16.04.
> However, the script does not work in Debian Stretch.
> The installation script tries to install `gcc-5` which does not exist in this distribution.

Cool, thanks. The gcc version can be configured in the script. I will try to make it work on Debain as well. Thanks for the hint.

butjar:
@jadinm Could you point me to the Debian and Fedora image you are testing against?

jadinm:
I use the vagrant boxes [bento/debian-9](https://app.vagrantup.com/bento/boxes/debian-9) and [bento/fedora-28](https://app.vagrantup.com/bento/boxes/fedora-28).

oliviertilmans:
> * OpenR requires 4GB of RAM to build (because it is a bit frustrating to discover it after an hour ^^)

Does it use LTO? The fact that it takes an hour to crash hints towards a link-time error. Disabling LTO/Enabling ThinLTO could help there (if LTO is used at all) at the expense of some slightly reduced perf. for openrd (nothing that matters here as we won't emulate networks of 1k nodes...).

butjar:
> > * OpenR requires 4GB of RAM to build (because it is a bit frustrating to discover it after an hour ^^)
> 
> Does it use LTO? The fact that it takes an hour to crash hints towards a link-time error. Disabling LTO/Enabling ThinLTO could help there (if LTO is used at all) at the expense of some slightly reduced perf. for openrd (nothing that matters here as we won't emulate networks of 1k nodes...).

I think it doesn't. However, I think the reason of the long build time results from more than 10 dependencies which are built from scratch as well. So there are several builds which could be probably optimized :disappointed:.

butjar:
I will adjust this PR to build OpenR for a Ubuntu VM only, but we can draw two issues from here:
- Speed up the OpenR installation
- Enhance OpenR installation for Fedora/Debian

util/install.py
butjar: :+1: I added it to the beginning of the function

butjar:
> I tested it. It works on Ubuntu 16.04.
> However, the script does not work in Debian Stretch.
> The installation script tries to install `gcc-5` which does not exist in this distribution.

Tested on Debian also with 'gcc-6', but fails at one of the dependencies.

butjar:
> I use the vagrant boxes [bento/debian-9](https://app.vagrantup.com/bento/boxes/debian-9) and [bento/fedora-28](https://app.vagrantup.com/bento/boxes/fedora-28).

If multi-OS portability is important https://kitchen.ci/ is also a neat tool (maybe combined with https://testinfra.readthedocs.io/en/latest/).

butjar:
Test TODOs:
- [ ] Test OpenR installation on Ubuntu (In progress)
- [ ] Test ipv6 enabling on Ubuntu
- [ ] Test ipv6 enabling on Debian
- [ ] Test ipv6 enabling on Fedora

butjar:
I tested the script on ubuntu, debian and fedora. Seems to work fine. However, fedora doesn't come with `python` pre-installed. Maybe we could adjust the shell script to prepare all distributions.

jadinm:
> I tested the script on ubuntu, debian and fedora. Seems to work fine.

Working on my side as well :+1: 

> However, fedora doesn't come with `python` pre-installed. Maybe we could adjust the shell script to prepare all distributions.

The problem is that Mininet only offers a Ubuntu VM boot script.
So I wouldn't do it as long as they don't support more.

PR #31, opened 2019-05-05
[WIP] Kitchen.ci test environment

butjar:
https://kitchen.ci/

Kitchen CI helps testing infrastructure code. Since ipmininet supports
various distributions kitchen.ci can help testing the provisioning
scripts.

### DIFF
--- a/kitchen.openr.yml
+++ b/kitchen.openr.yml
@@ -1,0 +1,19 @@
+---
+driver:
+  name: vagrant
+  synced_folders:
+    - [".", "/opt/ipmininet"]
+  customize:
+    memory: 4096
+
+provisioner:
+  name: shell
+  command: /opt/ipmininet/util/build_vm.sh
+
+platforms:
+  - name: ubuntu-16.04
+
+suites:
+  - name: default
+    run_list:
+    attributes:
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -1,0 +1,19 @@
+---
+driver:
+  name: vagrant
+  synced_folders:
+    - [".", "/opt/ipmininet"]
+
+provisioner:
+  name: shell
+  command: sudo python /opt/ipmininet/util/install.py -a
+
+platforms:
+  - name: ubuntu-16.04
+  - name: debian-9
+  - name: fedora-28
+
+suites:
+  - name: default
+    run_list:
+    attributes:
--- a/util/install.py
+++ b/util/install.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 import argparse
 import os
 import sys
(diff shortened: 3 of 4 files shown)


butjar:
If you wanna try you could try installing the ruby gem `gem install kitchen-vagrant`. However, for me installing `chefdk` from https://downloads.chef.io/chefdk always worked best.

butjar:
Currently there are at least two issues with that PR:
1. Fedora doesn't converge, because python is missing
2. `build_vm.sh` script gets somehow stuck

I just created the PR to show you the kitchen workflow so you can decide if it makes sense to add.

kitchen.openr.yml
```diff
@@ -0,0 +1,19 @@
...
+driver:
+  name: vagrant
+  synced_folders:
+    - [".", "/opt/ipmininet"]
+  customize:
+    memory: 4096
+
+provisioner:
+  name: shell
+  command: /opt/ipmininet/util/build_vm.sh
```
jadinm:
```suggestion
  command: /opt/ipmininet/util/build_vm.sh && sudo pytest -v /opt/ipmininet
```

kitchen.yml
```diff
@@ -0,0 +1,19 @@
+---
+driver:
+  name: vagrant
+  synced_folders:
+    - [".", "/opt/ipmininet"]
+
+provisioner:
+  name: shell
+  command: sudo python /opt/ipmininet/util/install.py -a
```
jadinm:
```suggestion
  command: if which apt; then sudo apt install -y python; elif which yum; then sudo yum install -y python; fi; sudo python /opt/ipmininet/util/install.py -iam && sudo pytest -v /opt/ipmininet
```

util/install.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
```
jadinm: It is not needed, I think
I would remove it

kitchen.yml
jadinm: This solves the problems for all platforms

kitchen.openr.yml
butjar: Kitchen supplies a config section (verifiers) for running tests https://kitchen.ci/docs/getting-started/writing-test/. Separating the provisioning and the tests. I suggest adding the tests there.

kitchen.yml
butjar: :+1: for `if which apt; then sudo apt install -y python; elif which yum; then sudo yum install -y python; fi; sudo python /opt/ipmininet/util/install.py -iam`

For `sudo pytest -v /opt/ipmininet` see https://github.com/cnp3/ipmininet/pull/31#discussion_r288594278

jadinm:
> This could make sense. I don't know how to solve the `build_vm.sh` stucked however.
> Maybe this is due to the updates of the grub which require user interactions.

If this is due to that, adding ```export DEBIAN_FRONTEND=noninteractive``` at the start of the build_vm.sh would fix the problem.

PR #32, opened 2019-06-12
[fix] Return host in IPNet addHost

butjar:
The addHost method of Mininet returns the host instance. Since we call
the method from the super class we should return the result.


jadinm:
Thanks for the PR!

PR #35, opened 2019-07-23
Add reathedoc documation

jadinm:
This PR consists mainly of sphinx configuration files, restructured text, and docstrings formating.

You can see how the doc looks on https://jadinm-ipmininet.readthedocs.io/en/doc/


oliviertilmans:
Nice!

Would there be a way to ensure that the doc snippets provided in doc/ are always 'working' and/or up-to-date? IIRC doctest can be used with sphinx to trigger tests from the doc comments but I am not sure on how easy that would be to use here...

jadinm:
It looks like it is possible with https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html

I made a small test with:

```rst
.. testcode::

    from ipmininet.iptopo import IPTopo

    class MyTopology(IPTopo):

        def build(self, *args, **kwargs):

            r1 = self.addRouter("r1")
            r2 = self.addRouter("r2")

            r1 = self.addSwitch("s1")
            s2 = self.addSwitch("s2")

            h1 = self.addHost("h1")
            h2 = self.addHost("h2")

            self.addLink(r1, r2)
            self.addLink(s1, r1)
            self.addLink(h1, s1)
            self.addLink(s2, r2)
            self.addLink(h2, s2)

            super(MyTopology, self).build(*args, **kwargs)

.. testcode::
    :hide:

    MyTopology().build()
```

Then a `make doctest` does the trick.

I will update the doc later

Issue #39, opened 2019-07-25
Pin mininet/... version


oliviertilmans:
The current setup.py script lists dependencies, but not their version.

It would make sense to explicitly pin the version used by ipmininet to ensure that future updates don't break things...

Out of all the dependencies, I think we should:
- Pin mininet to an exact version
- restrict mako to a major version
- specify a minimum version for everything else

Is mininet even on pypi at all? If not, we can add in the setup script how to install the particular version we require (so that issuing 'setup.py install' is the one-liner to get ipmininet running)

PR #41, opened 2019-07-26
Pins dependency version and fix mininet install with pip

jadinm:
Closes Issue #39 

This pins versions of dependencies in the `setup.py`.
It also installs the Mininet dependencies if not already present.

This implies changes in the install scripts: they are moved inside the ipmininet library so that they can be used by the `setup.py` file. The documentation is also updated.

Some points to discuss:

1. The chosen versions for each dependency
2. The Mininet dependencies (that are not packages) are installed in $HOME/mininet_dependencies. Maybe another location would be more suitable.
3. The syntax `mininet @ git+https://github.com/mininet/mininet@2.3.0d6` is only supported from pip 19. There is another way of installing a dependency from github but it is [deprecated](https://github.com/pypa/pip/issues/4187) and it requires extra arguments when running pip. Still, this might not be the best solution.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -46,6 +46,34 @@ following command in the same directory as the two previous one:
 .. _Vagrant: https://www.vagrantup.com/downloads.html
 .. _Virtualbox: https://www.virtualbox.org/wiki/Downloads
 
+Manual installation from PyPI
+-----------------------------
+
+You can download and install IPMininet.
+If you have pip above **18.1**, execute:
+
+.. code-block:: bash
+
+    $ sudo pip install ipmininet
+
+If you have an older version of pip, use:
+
+.. code-block:: bash
+
+    $ sudo pip install --process-dependency-links ipmininet
+
+Then, you can install all the daemons:
+
+.. code-block:: bash
+
+    $ sudo python -m ipmininet.install -af
+
+You can choose to install only a subset of the daemons
+by changing the options on the installation script.
+For the option documentations, use the ``-h`` option.
+
+.. _documentation: http://mininet.org/download/
+
 Manual installation from sources
 --------------------------------
 
@@ -57,35 +85,25 @@ To manually install IPMininet from source, first get the source code:
     $ cd ipmininet
     $ git checkout <version>
 
-Then, install IPMininet, Mininet and all the daemons:
+Then, install IPMininet.
+If you have pip above **18.1**, execute:
 
 .. code-block:: bash
 
-    $ sudo python util/install.py -iamf
+    $ sudo pip install .
 
-You can choose to install only a subset of the daemons
-by changing the options on the installation script.
-For the option documentations, use the ``-h`` option.
-
-Manual installation from PyPI
------------------------------
-
-Install Mininet by following its `documentation`_.
-
-Then, you can download and install IPMininet.
+If you have an older version of pip, use:
 
 .. code-block:: bash
 
-    $ sudo pip install ipmininet
+    $ sudo pip install --process-dependency-links .
 
 Finally, you can install all the daemons:
 
 .. code-block:: bash
 
-    $ sudo python util/install.py -af
+    $ sudo python -m ipmininet.install -af
 
 You can choose to install only a subset of the daemons
 by changing the options on the installation script.
 For the option documentations, use the ``-h`` option.
-
-.. _documentation: http://mininet.org/download/
--- a/ipmininet/install/__main__.py
+++ b/ipmininet/install/__main__.py
@@ -1,0 +1,56 @@
+from .install import *
+
+
+if __name__ == "__main__":
+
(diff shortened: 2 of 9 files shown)


oliviertilmans:
>     1. The chosen versions for each dependency

Special care needs to be taken with pinned deps (i.e., '==') so that ipmininet does not prevent those packages from being upgraded in the future...

>     2. The Mininet dependencies (that are not packages) are installed in $HOME/mininet_dependencies. Maybe another location would be more suitable.

I'd personally for for /opt for that, and then either symlinking the exec in /usr/bin or modifying $PATH (possibly using /etc/environment or ...)

>     3. The syntax `mininet @ git+https://github.com/mininet/mininet@2.3.0d6` is only supported from pip 19. There is another way of installing a dependency from github but it is [deprecated](https://github.com/pypa/pip/issues/4187) and it requires extra arguments when running pip. Still, this might not be the best solution.

I am certain I used a similar syntax without issues a few years ago (possibly with the now deprecated hook). An alternative would be to auto-detect whether mininet is installed in your install script and do it 'manually' (i.e., side-stepping pip as we're not using pypi for that dep. anyway).

jadinm:
> > ```
> > 1. The chosen versions for each dependency
> > ```
> 
> Special care needs to be taken with pinned deps (i.e., '==') so that ipmininet does not prevent those packages from being upgraded in the future...

I only set mininet with a strict version.
I chose 2.3.0d6 because 2.2.2 (the last stable version) does not support Python3.

> 
> > ```
> > 2. The Mininet dependencies (that are not packages) are installed in $HOME/mininet_dependencies. Maybe another location would be more suitable.
> > ```
> 
> I'd personally for for /opt for that, and then either symlinking the exec in /usr/bin or modifying $PATH (possibly using /etc/environment or ...)

Ok
 
> > ```
> > 3. The syntax `mininet @ git+https://github.com/mininet/mininet@2.3.0d6` is only supported from pip 19. There is another way of installing a dependency from github but it is [deprecated](https://github.com/pypa/pip/issues/4187) and it requires extra arguments when running pip. Still, this might not be the best solution.
> > ```
> 
> I am certain I used a similar syntax without issues a few years ago (possibly with the now deprecated hook). An alternative would be to auto-detect whether mininet is installed in your install script and do it 'manually' (i.e., side-stepping pip as we're not using pypi for that dep. anyway).

Yes, with the deprecated argument, you could set a dependency_links parameter with github links from pip 10. And actually, the git link directly in the install_require parameter works since pip 18.1.

I would rather not use the manual strategy. This might prevent someone from using specific parameters of pip (e.g., "--target \<dir\>") and it is also a bit strange not to have mininet as explicit python dependence.

oliviertilmans:
> I only set mininet with a strict version.
> I chose 2.3.0d6 because 2.2.2 (the last stable version) does not support
> Python3.

Sounds reasonable, I was mostly pointing to dependencies such as futures/...

> Yes, with the deprecated argument, you could set a dependency_links
> parameter with github links from pip 10. And actually, the git link directly in
> the install_require parameter works since pip 18.1.
> 
> I would rather not use the manual strategy. This might prevent someone from
> using specific parameters of pip (e.g., "--target <dir>") and it is also a bit
> strange not to have mininet as explicit python dependence.

Good point. I can you add version check in the setup script to warn that it will fail
to install mininet and maybe print a manual fallback to be executed by the user?
Beside updating pip itsel ofc (which might be managed by the distribution hence
lag behind a few versions). Such issue is temporary anyway and we may just ignore
it altogether past documentation.

jadinm:
> Good point. I can you add version check in the setup script to warn that it will fail
to install mininet and maybe print a manual fallback to be executed by the user?
Beside updating pip itsel ofc (which might be managed by the distribution hence
lag behind a few versions). Such issue is temporary anyway and we may just ignore
it altogether past documentation.

Ok, I modified the setup.py file to take into account the pip version.
I cannot check that users added the "--process-dependency-links" parameter because it is handled internally by pip and I don't have access to that.
So, I printed a warning (see below) but pip won't show it unless users add a "-v" to the pip call.

Nevertheless, this improves the PR since it is does no longer require to update pip.

```python
# Get back Pip version
try:
    version = parse_version(require("pip")[0].version)
except IndexError:
    version = parse_version("0")
    print("We cannot find the version of pip."
          "We assume that is it inferior to 18.1.")

if version >= parse_version("18.1"):
    install_requires.append('mininet @ git+https://github.com/mininet/mininet@{ver}'
                            .format(ver=MININET_VERSION))
else:
    print("You should run pip with --process-dependency-links to install all dependencies")
    install_requires.append('mininet=={ver}'.format(ver=MININET_VERSION))
    dependency_links.append('git+https://github.com/mininet/mininet@{ver}#egg=mininet-{ver}'
                            .format(ver=MININET_VERSION))
```

PR #59, opened 2019-11-09
Fix behavior with global loopback addresses

jadinm:
- Make the interface appear when calling router.intfList() by changing
  the port number to -1. Without that, it is erased from the list of router interfaces.
- Make OSPF/OSPF6/RIPng route these addresses
- Declare the interface in Zebra
- Allow CLI commands to use and recognize these addresses


jadinm:
Jenkins, run tests

jadinm:
Jenkins, rerun tests

PR #60, opened 2019-11-20
Fix broadcast domain building

jadinm:
In Python2, the items() method returns a list and not an iterator.
Therefore, changing the values of the dictionary in the loop had no
effect and a broadcast domain for each interface was created.


jadinm:
Jenkins, run tests.

PR #64, opened 2019-11-24
Fix CLI behavior with switches

jadinm:
Mininet Switches have a loopback interface intialized with a plain Intf
class and not the IPIntf class, Therefore it misses ips() and ip6s()
methods.
Moreover, the fact that switches do not have IP addresses, do not
have to prevent the user from running commands on them.


jadinm:
Jenkins, run tests.

PR #67, opened 2020-01-07
Remove support for Python 2 and add Python 3 features

jadinm:
This PR removes support for Python version strictly below Python 3.5 since it is the oldest supported version at the moment.


jadinm:
Jenkins, run tests.

jadinm:
Jenkins, rerun tests.

PR #75, opened 2020-06-21
Fix use_v4 and use_v6 impact on address allocation

### LINKED ISSUE #72: [IPv6-only or IPv4-only] Fix use_vX=False in router arguments
Passing arguments use_v6=False to addRouter prevents it to start a OSPFv3 daemon but it still assigns it an IPv6 global address

We would expect no IPv6 address allocated on the router

jadinm:
No address of a given IP version is allocated on an address whose node is not supposed to use.
This also adds an automated test.

Fix #72


jadinm:
Jenkins, run tests.

PR #76, opened 2020-06-23
Fix openr install

butjar:
The build script previously generated from the OpenR repository didn't fix all dependencies properly. Therefore the OpenR build in the install scripts was broken. This PR starts from a freshly generated build script and contains commits that pin dependencies manually. Additionally, the OpenR python dependencies can be installed with either python2 or python3.

The vm build script (with OpenR installation) has been tested under Ubuntu 18.04.

### DIFF
--- a/ipmininet/install/build_openr-rc-20190419-11514.sh
+++ b/ipmininet/install/build_openr-rc-20190419-11514.sh
@@ -1,0 +1,281 @@
+#!/bin/bash
+
+set -exo pipefail
+### Install packages for Debian-based OS ###
+
+apt-get update && apt-get install -yq autoconf-archive bison build-essential cmake curl flex git gperf joe libboost-all-dev libcap-dev libdouble-conversion-dev libevent-dev libgflags-dev libgoogle-glog-dev libkrb5-dev libpcre3-dev libpthread-stubs0-dev libnuma-dev libsasl2-dev libsnappy-dev libsqlite3-dev libssl-dev libtool netcat-openbsd pkg-config sudo unzip wget python3-venv python-setuptools python3-setuptools python-pip ccache
+apt-get install -yq gcc-'5' g++-'5'
+update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-'5' 40 --slave /usr/bin/g++ g++ /usr/bin/g++-'5'
+update-alternatives --config gcc
+
+export CCACHE_DIR='/ccache' CC="ccache ${CC:-gcc}" CXX="ccache ${CXX:-g++}"
+### Diagnostics ###
+
+# Builder DebianSystemFBCodeBuilder(google/googletest:cmake_defines={u'BUILD_GTEST': u'ON', u'BUILD_SHARED_LIBS': u'OFF'}, gcc_version=u'5', facebook/zstd:git_hash=ShellQuoted(u'$(git describe --abbrev=0 --tags origin/master)'), hyperic/sigar:autoconf_options={u'CFLAGS': u'-fgnu89-inline'}, openr/build:cmake_defines={u'ADD_ROOT_TESTS': u'OFF'}, prefix=u'/usr/local', projects_dir=u'/usr/local/src', google/googletest:git_hash=u'release-1.8.1', wangle/wangle/build:cmake_defines={u'BUILD_TESTS': u'OFF'}, thom311/libnl:git_hash=u'libnl3_2_25', jedisct1/libsodium:git_hash=u'stable', ccache_dir=u'/ccache', make_parallelism=4, no1msd/mstch:git_hash=ShellQuoted(u'$(git describe --abbrev=0 --tags)'), zeromq/libzmq:git_hash=u'v4.2.5')
+hostname
+cat /etc/issue || echo no /etc/issue
+g++ --version || echo g++ not installed
+cmake --version || echo cmake not installed
+
+### Check out facebook/folly, workdir _build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'folly' ]]; then 
+	git clone https://github.com/'facebook/folly'
+fi
+mkdir -p '/usr/local/src'/'folly'/'_build' && cd '/usr/local/src'/'folly'/'_build'
+git checkout '3ceffd7d145be3c85a2aae39f99eb86ea730bdcc'
+
+### Build and install facebook/folly ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out jedisct1/libsodium, workdir . ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'libsodium' ]]; then 
+	git clone https://github.com/'jedisct1/libsodium'
+fi
+mkdir -p '/usr/local/src'/'libsodium'/'.' && cd '/usr/local/src'/'libsodium'/'.'
+git checkout 'stable'
+
+### Build and install jedisct1/libsodium ###
+
+./autogen.sh
+LDFLAGS="$LDFLAGS" CFLAGS="$CFLAGS" CPPFLAGS="$CPPFLAGS" ./configure 
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out facebookincubator/fizz, workdir fizz/build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'fizz' ]]; then 
+	git clone https://github.com/'facebookincubator/fizz'
+fi
+mkdir -p '/usr/local/src'/'fizz'/'fizz/build' && cd '/usr/local/src'/'fizz'/'fizz/build'
+
+### Build and install fizz/fizz/build ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out google/googletest, workdir build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'googletest' ]]; then 
+	git clone https://github.com/'google/googletest'
+fi
+mkdir -p '/usr/local/src'/'googletest'/'build' && cd '/usr/local/src'/'googletest'/'build'
+git checkout 'release-1.8.1'
+
+### Build and install google/googletest ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_GTEST'='ON' -D'BUILD_SHARED_LIBS'='OFF' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out rsocket/rsocket-cpp, workdir rsocket ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'rsocket-cpp' ]]; then 
+	git clone https://github.com/'rsocket/rsocket-cpp'
+fi
+mkdir -p '/usr/local/src'/'rsocket-cpp'/'rsocket' && cd '/usr/local/src'/'rsocket-cpp'/'rsocket'
+
+### Build and install rsocket-cpp/rsocket ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out facebook/wangle, workdir wangle/build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'wangle' ]]; then 
+	git clone https://github.com/'facebook/wangle'
+fi
+mkdir -p '/usr/local/src'/'wangle'/'wangle/build' && cd '/usr/local/src'/'wangle'/'wangle/build'
+git checkout 'f2f8e1996739df1fc8ab1e003a1bc6472bd5b9bd'
+
+### Build and install wangle/wangle/build ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' -D'BUILD_TESTS'='OFF' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out facebook/zstd, workdir . ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'zstd' ]]; then 
+	git clone https://github.com/'facebook/zstd'
+fi
+mkdir -p '/usr/local/src'/'zstd'/'.' && cd '/usr/local/src'/'zstd'/'.'
+git checkout $(git describe --abbrev=0 --tags origin/master)
+
+### Build and install zstd ###
+
+make -j '4' VERBOSE=1 'PREFIX'='/usr/local'
+sudo make install VERBOSE=1 'PREFIX'='/usr/local'
+sudo ldconfig
+
+### Check out no1msd/mstch, workdir build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'mstch' ]]; then 
+	git clone https://github.com/'no1msd/mstch'
+fi
+mkdir -p '/usr/local/src'/'mstch'/'build' && cd '/usr/local/src'/'mstch'/'build'
+git checkout $(git describe --abbrev=0 --tags)
+
+### Build and install no1msd/mstch ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out facebook/fbthrift, workdir thrift ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'fbthrift' ]]; then 
+	git clone https://github.com/'facebook/fbthrift'
+fi
+mkdir -p '/usr/local/src'/'fbthrift'/'thrift' && cd '/usr/local/src'/'fbthrift'/'thrift'
+git checkout '2e3a85eb29e0cc3bad910093656bc2196f9e96ae'
+
+### Build and install fbthrift/thrift ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Install thrift python modules ###
+
+mkdir -p '/usr/local/src'/'fbthrift/thrift/lib/py' && cd '/usr/local/src'/'fbthrift/thrift/lib/py'
+sudo python setup.py install
+
+### Check out hyperic/sigar, workdir . ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'sigar' ]]; then 
+	git clone https://github.com/'hyperic/sigar'
+fi
+mkdir -p '/usr/local/src'/'sigar'/'.' && cd '/usr/local/src'/'sigar'/'.'
+
+### Build and install sigar ###
+
+./autogen.sh
+LDFLAGS="$LDFLAGS" CFLAGS="$CFLAGS" CPPFLAGS="$CPPFLAGS" ./configure 'CFLAGS'='-fgnu89-inline'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out zeromq/libzmq, workdir . ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'libzmq' ]]; then 
+	git clone https://github.com/'zeromq/libzmq'
+fi
+mkdir -p '/usr/local/src'/'libzmq'/'.' && cd '/usr/local/src'/'libzmq'/'.'
+git checkout 'v4.2.5'
+
+### Build and install zeromq/libzmq ###
+
+./autogen.sh
+LDFLAGS="$LDFLAGS" CFLAGS="$CFLAGS" CPPFLAGS="$CPPFLAGS" ./configure 
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out facebook/fbzmq, workdir fbzmq/build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'fbzmq' ]]; then 
+	git clone https://github.com/'facebook/fbzmq'
+fi
+mkdir -p '/usr/local/src'/'fbzmq'/'fbzmq/build' && cd '/usr/local/src'/'fbzmq'/'fbzmq/build'
+git checkout '8fba3b727c8194031351cefee71bd36ba3486645'
+
+### Build and install fbzmq/fbzmq/build ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+PYTHONPATH="$PYTHONPATH:"'/usr/local'/lib/python2.7/site-packages make -j '4'
+sudo make install
+sudo ldconfig
+
+### Install fbzmq python modules ###
+
+mkdir -p '/usr/local/src'/'fbzmq/fbzmq/py' && cd '/usr/local/src'/'fbzmq/fbzmq/py'
+sudo python setup.py install
+
+### Check out google/re2, workdir build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'re2' ]]; then 
+	git clone https://github.com/'google/re2'
+fi
+mkdir -p '/usr/local/src'/'re2'/'build' && cd '/usr/local/src'/'re2'/'build'
+
+### Build and install google/re2 ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out thom311/libnl, workdir . ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'libnl' ]]; then 
+	git clone https://github.com/'thom311/libnl'
+fi
+mkdir -p '/usr/local/src'/'libnl'/'.' && cd '/usr/local/src'/'libnl'/'.'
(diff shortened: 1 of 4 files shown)


ipmininet/install/build_vm.sh
```diff
@@ -30,8 +33,9 @@ sudo pip3 install --upgrade pip
 sudo apt remove -yq python3-pip
 
 # Install ipmininet
-git clone https://github.com/cnp3/ipmininet.git
+[ ! -d "ipmininet" ] && git clone ${IPMN_REPO}
```
oliviertilmans, line 36: This is a great idea, could take this one step further and add a variable for the clone target?
E.g.
```suggestion
IPMN_DIR="${IPMN_DIR:-ipmininet}"
[ ! -d "${IPMN_DIR" ] && git clone ${IPMN_REPO}${IPMN_DIR}
pushd ${IPMN_DIR}
```

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
+#!/bin/bash
```
oliviertilmans, line 1: Did you write this install script yourself, or was it generated by some of the OpenR build-tools?

How "portable" across OpenR release should this be?

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+
+set -exo pipefail
+### Install packages for Debian-based OS ###
+
+apt-get update && apt-get install -yq autoconf-archive bison build-essential cmake curl flex git gperf joe libboost-all-dev libcap-dev libdouble-conversion-dev libevent-dev libgflags-dev libgoogle-glog-dev libkrb5-dev libpcre3-dev libpthread-stubs0-dev libnuma-dev libsasl2-dev libsnappy-dev libsqlite3-dev libssl-dev libtool netcat-openbsd pkg-config sudo unzip wget python3-venv python-setuptools python3-setuptools python-pip python3-pip ccache
```
oliviertilmans, line 6: IIRC the install script does not guarantee that the host system is running debian/has apt-get and update-alternatives.
You should thus either disable OpenR (with a warning) on non-debian (likely on the python install script); or support the alternatives here.

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+apt-get install -yq gcc-'5' g++-'5'
+update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-'5' 40 --slave /usr/bin/g++ g++ /usr/bin/g++-'5'
+update-alternatives --config gcc
+
+PY_MAJOR=${PY_MAJOR:-2}
+
+if [ "$PY_MAJOR" -ne 2 ] && [ "$PY_MAJOR" -ne 3 ]; then
```
oliviertilmans, line 13: I would not bother with PY_MAJOR=2, as ipmininet dropped support for python2

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+echo "PY_LIB_PATH: $PY_LIB_PATH"
+
+export CCACHE_DIR='/ccache' CC="ccache ${CC:-gcc}" CXX="ccache ${CXX:-g++}"
+### Diagnostics ###
+
+# Builder DebianSystemFBCodeBuilder(google/googletest:cmake_defines={u'BUILD_GTEST': u'ON', u'BUILD_SHARED_LIBS': u'OFF'}, gcc_version=u'5', facebook/zstd:git_hash=ShellQuoted(u'$(git describe --abbrev=0 --tags origin/master)'), hyperic/sigar:autoconf_options={u'CFLAGS': u'-fgnu89-inline'}, openr/build:cmake_defines={u'ADD_ROOT_TESTS': u'OFF'}, prefix=u'/usr/local', projects_dir=u'/usr/local/src', google/googletest:git_hash=u'release-1.8.1', wangle/wangle/build:cmake_defines={u'BUILD_TESTS': u'OFF'}, thom311/libnl:git_hash=u'libnl3_2_25', jedisct1/libsodium:git_hash=u'stable', ccache_dir=u'/ccache', make_parallelism=4, no1msd/mstch:git_hash=ShellQuoted(u'$(git describe --abbrev=0 --tags)'), zeromq/libzmq:git_hash=u'v4.2.5')
+hostname
+cat /etc/issue || echo no /etc/issue
+g++ --version || echo g++ not installed
+cmake --version || echo cmake not installed
```
oliviertilmans, line 33: Shouldn't these cause the exit of the script?

ipmininet/install/install.py
```diff
@@ -152,27 +151,18 @@ def install_frrouting(output_dir: str):
...
+def install_openr(output_dir: str, py_major="3"):
+    # It's not possible to get a build script with pinned dependencies from the
+    # OpenR github repository. The checked-in build script has the dependencies
+    # pinned manually. Builds and installs OpenR release rc-20190419-11514.
+    # https://github.com/facebook/openr/releases/tag/rc-20190419-11514
+    script_name = "build_openr-rc-20190419-11514.sh"
+    openr_buildscript = os.path.join("ipmininet/ipmininet/install/",
+                                     script_name)
     # Execute build script
     sh(openr_buildscript,
```
oliviertilmans, line 163: Given that the script can fail, you should check the return code after this.

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+
+### Check out facebook/folly, workdir _build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'folly' ]]; then 
+	git clone https://github.com/'facebook/folly'
+fi
+mkdir -p '/usr/local/src'/'folly'/'_build' && cd '/usr/local/src'/'folly'/'_build'
+git checkout '3ceffd7d145be3c85a2aae39f99eb86ea730bdcc'
```
oliviertilmans, line 42: I assume that this particular commit corresponds to the openR tag you're building.
Could you put this in a top-level variable, in the hope that upgrading OpenR releases can then be simplified by changing the top-level $FOLLY_REV, &co?

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+### Build and install facebook/folly ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+make -j '4' VERBOSE=1 
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out jedisct1/libsodium, workdir . ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
```
oliviertilmans, line 53: You've already checked for the existence of that dir (and created it above).

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+if [[ ! -d '/usr/local/src'/'libsodium' ]]; then 
+	git clone https://github.com/'jedisct1/libsodium'
+fi
+mkdir -p '/usr/local/src'/'libsodium'/'.' && cd '/usr/local/src'/'libsodium'/'.'
```
oliviertilmans, line 57: Given you've cloned the source in '/usr/local/src'/'libsodium', creating a dir at '/usr/local/src'/'libsodium/.' achieves nothing.

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+
+PY_VERSION="$($PY_BIN --version 2>&1)"
+PY_MINOR=$(sed -e 's/\(Python \)\([[:digit:]]\)\.\([[:digit:]]\)\(.*\)/\3/g' <<< "$PY_VERSION")
+PY_LIB_PATH="/usr/local/lib/python$PY_MAJOR.$PY_MINOR/site-packages"
+echo "PY_LIB_PATH: $PY_LIB_PATH"
+
+export CCACHE_DIR='/ccache' CC="ccache ${CC:-gcc}" CXX="ccache ${CXX:-g++}"
+### Diagnostics ###
+
+# Builder DebianSystemFBCodeBuilder(google/googletest:cmake_defines={u'BUILD_GTEST': u'ON', u'BUILD_SHARED_LIBS': u'OFF'}, gcc_version=u'5', facebook/zstd:git_hash=ShellQuoted(u'$(git describe --abbrev=0 --tags origin/master)'), hyperic/sigar:autoconf_options={u'CFLAGS': u'-fgnu89-inline'}, openr/build:cmake_defines={u'ADD_ROOT_TESTS': u'OFF'}, prefix=u'/usr/local', projects_dir=u'/usr/local/src', google/googletest:git_hash=u'release-1.8.1', wangle/wangle/build:cmake_defines={u'BUILD_TESTS': u'OFF'}, thom311/libnl:git_hash=u'libnl3_2_25', jedisct1/libsodium:git_hash=u'stable', ccache_dir=u'/ccache', make_parallelism=4, no1msd/mstch:git_hash=ShellQuoted(u'$(git describe --abbrev=0 --tags)'), zeromq/libzmq:git_hash=u'v4.2.5')
```
oliviertilmans, line 29: What's the purpose of this line?
Replacing everything that is below if running on GCE? If so, you could use it as-is with some env detection?

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+sudo make install VERBOSE=1 
+sudo ldconfig
+
+### Check out facebookincubator/fizz, workdir fizz/build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'fizz' ]]; then 
+	git clone https://github.com/'facebookincubator/fizz'
+fi
+mkdir -p '/usr/local/src'/'fizz'/'fizz/build' && cd '/usr/local/src'/'fizz'/'fizz/build'
```
oliviertilmans, line 74: This seems wrong, as you're creating a build subdir in fizz/fizz/build, whereas that repo already has one at fizz/build

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+g++ --version || echo g++ not installed
+cmake --version || echo cmake not installed
+
+### Check out facebook/folly, workdir _build ###
+
+mkdir -p '/usr/local/src' && cd '/usr/local/src'
+if [[ ! -d '/usr/local/src'/'folly' ]]; then 
+	git clone https://github.com/'facebook/folly'
+fi
+mkdir -p '/usr/local/src'/'folly'/'_build' && cd '/usr/local/src'/'folly'/'_build'
```
oliviertilmans, line 41: Could you try to get rid of the extra ' ? e.g.,
```suggestion
mkdir -p '/usr/local/src/folly/_build' && cd '/usr/local/src/folly/_build'
```

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+git checkout '3ceffd7d145be3c85a2aae39f99eb86ea730bdcc'
+
+### Build and install facebook/folly ###
+
+CXXFLAGS="$CXXFLAGS -fPIC" CFLAGS="$CFLAGS -fPIC" cmake -D'BUILD_SHARED_LIBS'='ON' '..'
+make -j '4' VERBOSE=1
```
oliviertilmans, line 47: Don't make asumptions on the cpu#, e.g., at the very least do
```suggestion
make -j "$(nproc)" VERBOSE=1 
```

A better solution could be to define at the top:
```bash
function _make()
{
    make -j "$(nproc)" "$@"
}
```

and then use it in the later invocations as

```suggestion
_make VERBOSE=1 
```

ipmininet/install/build_openr-rc-20190419-11514.sh
```diff
@@ -0,0 +1,301 @@
...
+if [[ ! -d '/usr/local/src'/'libsodium' ]]; then 
+	git clone https://github.com/'jedisct1/libsodium'
+fi
+mkdir -p '/usr/local/src'/'libsodium'/'.' && cd '/usr/local/src'/'libsodium'/'.'
+git checkout '6bece9c8c45259998f83ce243b1933e76c03f545'
+
+### Build and install jedisct1/libsodium ###
+
+./autogen.sh
+LDFLAGS="$LDFLAGS" CFLAGS="$CFLAGS" CPPFLAGS="$CPPFLAGS" ./configure
```
oliviertilmans, line 63:
```suggestion
./configure 
```

ipmininet/install/build_vm.sh
butjar: > What's preventing us from directly following the steps from the OpenR repo?
I.e., https://github.com/facebook/openr/tree/master/build#build-on-debian-host-without-docker

Main problem is that the generated build scripts do not pin dependencies properly. My idea was to adjust the generated script and pin dependencies to create a stable build script for the 'ipmininet supported' release.

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: I generated the basic script from https://github.com/facebook/openr/tree/master/build#build-on-debian-host-without-docker.
The 'vanilla' script can be found in 622f6109bda4b91fd76adb577c175fef181182d6.
The following manual adjustments to the build script mainly pin dependencies for the release.

I would like to make it more portable. However, it seems that there is some re-work going on for the build chain and the documented build process is very brittle at the moment. I'm keeping an eye on the development and I will adopt the 'official' build instructions as soon as they have been stabilized.

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: The OpenR install scripts in ipmininet currently support Debian/Ubuntu only:
https://github.com/cnp3/ipmininet/blob/master/ipmininet/install/__main__.py#L55

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: OpenR uses python afaik for `breeze`, only. `Breeze` is a CLI tool that helps to assert the state of the running OpenR deamon: https://github.com/facebook/openr/blob/master/openr/docs/Breeze.md. Installing `breeze` with Python 2 has no effect on ipmininet. 

However, I guess your suggestion is to remove python 2.x also from "ipmininet-OpenR", right?

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: Are you asking about ccache or the comment?

- The former is a compiler cache that speeds up recompilation (I'm also not familiar with it).
- The latter is a comment with diagnostic purpose only, I guess. I belief the intention is to help you debugging the generated build script.

butjar:
I think all your suggestions to the generated build script make sense and I can adjust the build script with your review to make the script more stable (this might be the main goal here).

I didn't spend too much effort in refactoring the generated script as I expected it to be subject to chance at some point. My intention was to stay close to the generated script to keep changes easy to apply to a newly generated script.

What approach do you prefer?

ipmininet/install/install.py
butjar: :+1:

ipmininet/install/build_vm.sh
butjar: :+1: for the change

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: I will move all shas to top-level envs :+1:

oliviertilmans:
> OpenR uses python afaik for breeze, only. Breeze is a CLI tool that helps
 > to assert the state of the running OpenR deamon:
 > https://github.com/facebook/openr/blob/master/openr/docs/Breeze.md.
 > Installing breeze with Python 2 has no effect on ipmininet.
 > 
 > However, I guess your suggestion is to remove python 2.x also from
 > "ipmininet-OpenR", right?

Yes. If python3 is not available, ipmininet won't run. Arguably this makes
that whole check redundant (in the ipmininet context).

oliviertilmans:
> The OpenR install scripts in ipmininet currently support Debian/Ubuntu
 > only:
 > https://github.com/cnp3/ipmininet/blob/master/ipmininet/install/__main__.py
 > #L55

I somehow missed that. Thanks!

ipmininet/install/build_openr-rc-20190419-11514.sh
oliviertilmans: Yes. If python3 is not available, ipmininet won't run. Arguably this makes
that whole check redundant (in the ipmininet context).

ipmininet/install/build_openr-rc-20190419-11514.sh
oliviertilmans: I somehow missed that. Thanks!

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: You are right, the cores are evaluated on script generation, I guess. What do you think about:
```
#!/bin/bash

MAKEFLAGS+="-j $(shell nproc)"
```

ipmininet/install/build_openr-rc-20190419-11514.sh
oliviertilmans: About the comment. Re-reading, I confused myself thinking it was something else.

oliviertilmans:
I had not fully understood that the whole script had been first auto-generated then patched by you.
Knowing this, do not spend time on refactoring anything that was auto-generated unless it is 'wrong' (e.g., the hardcoded -j4 for make).

If you want to be more long-term focused, you could split the script into:
- The generated base
- a patch file (which you can extract by diffing the current one with the base)
Then add a step in the install script to generate the patched build scipt?

This might ease up future releases...

But that is also more work, and I understand it might not be worth it if the build-system changes in the near future

ipmininet/install/build_openr-rc-20190419-11514.sh
oliviertilmans: That's a good workaround as well yes.

ipmininet/install/build_openr-rc-20190419-11514.sh
oliviertilmans: Given that g++ and cmake are installed at the start of the script, these should never fail anyway.

ipmininet/install/build_vm.sh
butjar: See 6fd96f1 (I added a missing `}` and ` ` to your suggestion)

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: Agree. See ae27c71

ipmininet/install/install.py
butjar: Basically, the `sh` command does not allow failing "recipies". However, I added few lines for clarification. I still have to test my solution. See f3315f8.

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: See: e792c3c

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: Does not make sense. As discussed, I would keep it like it was generated.

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: Does not make sense. As discussed, I would keep it like it was generated.

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: Does not make sense. As discussed, I would keep it like it was generated.

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: Does not make sense. As discussed, I would keep it like it was generated.

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: See: b7acf86

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: Does not make sense. As discussed, I would keep it like it was generated.

butjar:
I worked in most of your suggestions, thx for the review. I'm currently testing the build scripts and let you know if it works like I expect. Please check if you can resolve the requested changes.

butjar:
> I had not fully understood that the whole script had been first auto-generated then patched by you.
> Knowing this, do not spend time on refactoring anything that was auto-generated unless it is 'wrong' (e.g., the hardcoded -j4 for make).
> 
> If you want to be more long-term focused, you could split the script into:
> 
> * The generated base
> * a patch file (which you can extract by diffing the current one with the base)
> 
> Then add a step in the install script to generate the patched build scipt?
> 
> This might ease up future releases...
> 
> But that is also more work, and I understand it might not be worth it if the build-system changes in the near future

Totally agree, I hope they FB revises their build chain soon and I can turn in a more appropriate solution. I keep your long-term solution in mind.

ipmininet/install/build_openr-rc-20190419-11514.sh
butjar: Log from my 2 core VM:
```
...
    terranet-base: ++ nproc
    terranet-base: + export 'MAKEFLAGS= -j 2'
    terranet-base: + MAKEFLAGS=' -j 2'
...
```

oliviertilmans:
I'll let @jadinm comment further and/or merge given he's the one that's actively looking after this project (:

butjar:
@jadinm @oliviertilmans I tested the `IPMN_DIR` option and the option to ignore failed OpenR build scripts here:
https://github.com/butjar/ipmininet/commit/afcaeb3d19c991975be3ec2f02afcd68d889a237

I had to pass down the checkout folder to the python install scripts: 71fe3d0

Seems to work fine. Output from my test VM:

```
...
    terranet-base: *** ipmn/ipmininet/install/build_openr-rc-20190419-11514.sh
    terranet-base: WARNING: Ignoring failed OpenR installation.
    terranet-base: ~ /usr/local/src
... 
```
(_installation continues_)

butjar:
I ran the script several times these days and it seems pretty stable to me.

butjar:
> Overall, it looks good to me.
> 
> One problem, you don't export the openr file when ipmininet is installed with pip.
> You either need to add it to MANIFEST.in (in which case it will stay at the same place in the exported code) or with the script parameter. First, let's show the manifest solution:
> 
> ```
> [...]
> include ipmininet/install/*.sh
> ```
> 
> You could then obtain this information without requiring a new argument with something like
> 
> ```python
>     openr_buildscript = os.path.join(os.path.dirname(os.path.abspath(__file__)), script_name)
> ```
> 
> The option is to use the "script" argument of the setup.py file (just as mininet is doing on its setup file, see https://github.com/mininet/mininet/blob/master/setup.py). In that case, the script is installed in the path. You could call it only by its name and it doesn't require the dir_name parameter.

Good point. I didn't put any thoughts into the packaging.

> ```python
>     openr_buildscript = os.path.join(os.path.dirname(os.path.abspath(__file__)), script_name)
> ```

:+1: Much more elegant.

My thoughts on MANIFEST.in vs script setup (vs data_files):
- MANIFEST.in + script path from `__file__` should work even without installing the package (e.g. if repo was just checked out and the install script is run). Right ?!
- script setup is actually thought for python scripts (See: https://docs.python.org/3/distutils/setupscript.html#installing-scripts)
- data_files seems more appropriate than scripts (See: https://docs.python.org/3/distutils/setupscript.html#installing-additional-files)

I prefer your proposed solution, which one do you prefer? Any other thoughts on this?

butjar:
> ```
> [...]
> include ipmininet/install/*.sh
> ```

Do we also want to include `build_vm.sh` though?

butjar:
@jadinm Your proposed solution seems to work fine:

```
    terranet-base:
    terranet-base: *** /root/ipmn/ipmininet/install/build_openr-rc-20190419-11514.sh
    terranet-base: + apt-get update
```
I removed the commit with the root directory passed to the install scripts with your proposed solution: 797e7ec355c4e2760300a34c4bde5401da52d7fc

Please check if you are fine with the new changes. I didn't test the packaging, can you take over for this test?

jadinm:
> Do we also want to include build_vm.sh though?

The idea is to include future versions of the openr script if we change the version in the future.
I think that it does not really matter if the build_vm.sh is included or not.

> I didn't test the packaging, can you take over and test it?

The packaging works

However, I ran into this error:
```
[ 53%] Building CXX object CMakeFiles/tckserver.dir/test/test_utils/StatsPrinter.cpp.o
/usr/bin/ccache  g++  -DDEBUG -DGFLAGS_IS_A_DLL=0 -I/usr/local/src/rsocket-cpp/rsocket/gmock-prefix/src/gmock/googlemock/include -I/usr/local/src/rsocket-cpp/rsocket/gmock-prefix/src/gmock/googletest/include -I/usr/local/src/rsocket-cpp -I/usr/local/src/rsocket-cpp/yarpl/.. -isystem /usr/local/include  -fPIC -Wall -Wextra -Woverloaded-virtual -g -Wno-unreachable-code -g   -std=gnu++14 -o CMakeFiles/tckserver.dir/test/test_utils/StatsPrinter.cpp.o -c /usr/local/src/rsocket-cpp/rsocket/test/test_utils/StatsPrinter.cpp
make[2]: *** No rule to make target 'gmock-prefix/src/gmock-build/./googlemock/libgmock.a', needed by 'tckserver'.  Stop.
make[2]: Leaving directory '/usr/local/src/rsocket-cpp/rsocket'
CMakeFiles/Makefile2:958: recipe for target 'CMakeFiles/tckserver.dir/all' failed
make[1]: *** [CMakeFiles/tckserver.dir/all] Error 2
make[1]: Leaving directory '/usr/local/src/rsocket-cpp/rsocket'
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
```
I am using a fresh Ubuntu 18.04

butjar:
> > Do we also want to include build_vm.sh though?
> 
> The idea is to include future versions of the openr script if we change the version in the future.
> I think that it does not really matter if the build_vm.sh is included or not.
>

:+1: agree.
 
> > I didn't test the packaging, can you take over and test it?
> 
> The packaging works
> 

Cool.

> However, I ran into this error:
> 
> ```
> [ 53%] Building CXX object CMakeFiles/tckserver.dir/test/test_utils/StatsPrinter.cpp.o
> /usr/bin/ccache  g++  -DDEBUG -DGFLAGS_IS_A_DLL=0 -I/usr/local/src/rsocket-cpp/rsocket/gmock-prefix/src/gmock/googlemock/include -I/usr/local/src/rsocket-cpp/rsocket/gmock-prefix/src/gmock/googletest/include -I/usr/local/src/rsocket-cpp -I/usr/local/src/rsocket-cpp/yarpl/.. -isystem /usr/local/include  -fPIC -Wall -Wextra -Woverloaded-virtual -g -Wno-unreachable-code -g   -std=gnu++14 -o CMakeFiles/tckserver.dir/test/test_utils/StatsPrinter.cpp.o -c /usr/local/src/rsocket-cpp/rsocket/test/test_utils/StatsPrinter.cpp
> make[2]: *** No rule to make target 'gmock-prefix/src/gmock-build/./googlemock/libgmock.a', needed by 'tckserver'.  Stop.
> make[2]: Leaving directory '/usr/local/src/rsocket-cpp/rsocket'
> CMakeFiles/Makefile2:958: recipe for target 'CMakeFiles/tckserver.dir/all' failed
> make[1]: *** [CMakeFiles/tckserver.dir/all] Error 2
> make[1]: Leaving directory '/usr/local/src/rsocket-cpp/rsocket'
> Makefile:140: recipe for target 'all' failed
> make: *** [all] Error 2
> ```
> 
> I am using a fresh Ubuntu 18.04

I can reproduce this issue. I tested the script on a ubuntu 18.04 vagrant box before (https://app.vagrantup.com/ubuntu/boxes/bionic64), where everything works fine. It's strange, because the build toolchain looks pretty similar on the first glance. I'm looking into it, thx for testing.

butjar:
@jadinm It seems to me that the problem arises when building on a single core. Can you can confirm that your build was triggered on a single core VM?

However, the root of the issue is probably, that the build dependencies are not properly defined. In case of `rsocket` the `gmock` dependency has not been built before the tests that require it. I ran into similar issues during the `openr` build, where some included files have not been generated yet. Interestingly these issues did not arise on a 2 core VM.

In e3ffeb90071b454b027d0d66742f5be18efb57a5 I trigger the build process of the required dependencies "manually".

jadinm:
>  @jadinm It seems to me that the problem arises when building on a single core. Can you can confirm that your build was triggered on a single core VM?

Yes, I used a single core VM

> However, the root of the issue is probably, that the build dependencies are not properly defined. In case of rsocket the gmock dependency has not been built before the tests that require it. I ran into similar issues during the openr build, where some included files have not been generated yet. Interestingly these issues did not arise on a 2 core VM.
>
> In e3ffeb9 I trigger the build process of the required dependencies "manually".

What an interesting side effect x)
I confirm that it now works on my side :+1:

butjar:
:metal:

PR #77, opened 2020-07-06
IPTopo: addLinks convenience helper

### LINKED ISSUE #81: Add addRouters QoL helper
Same idea than #77

oliviertilmans:
This commits adds a small QoL helper which trims down the addLink
boilerplate.

Also fixes #81

### DIFF
--- a/ipmininet/iptopo.py
+++ b/ipmininet/iptopo.py
@@ -89,6 +89,14 @@ class IPTopo(Topo):
                                            self.linkInfo(node1, node2, key))
         return link_description
 
+    def addLinks(self, *links, **opts):
+        """Add several links in one go.
+
+        :param links: link description tuples
+        :param opts: link otions (optional)"""
+        for descr in links:
+            self.addLink(*descr, **opts)
+
     def addDaemon(self, node: str, daemon: Union[Daemon, Type[Daemon]],
                   default_cfg_class: Type[NodeConfig] = BasicRouterConfig,
                   cfg_daemon_list="daemons", **daemon_params):
--- a/ipmininet/iptopo.py
+++ b/ipmininet/iptopo.py
@@ -89,13 +89,12 @@ class IPTopo(Topo):
                                            self.linkInfo(node1, node2, key))
         return link_description
 
-    def addLinks(self, *links, **opts):
+    def addLinks(self, *links, **opts) -> List['LinkDescription']:
         """Add several links in one go.
 
         :param links: link description tuples
-        :param opts: link otions (optional)"""
-        for descr in links:
-            self.addLink(*descr, **opts)
+        :param opts: link options (optional)"""
+        return [self.addLink(*descr, **opts) for descr in links]
 
     def addDaemon(self, node: str, daemon: Union[Daemon, Type[Daemon]],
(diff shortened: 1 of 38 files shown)


ipmininet/iptopo.py
```diff
@@ -89,6 +89,14 @@ def addLink(self, node1: str, node2: str, port1=None, port2=None,
                                            self.linkInfo(node1, node2, key))
         return link_description
 
+    def addLinks(self, *links, **opts):
+        """Add several links in one go.
+
+        :param links: link description tuples
+        :param opts: link otions (optional)"""
+        for descr in links:
```
jadinm, line 97:
```suggestion

```

ipmininet/iptopo.py
```diff
@@ -89,6 +89,14 @@ def addLink(self, node1: str, node2: str, port1=None, port2=None,
...
+            self.addLink(*descr, **opts)
```
jadinm, line 98:
```suggestion
        return [self.addLink(*descr, **opts) for descr in links]
```

oliviertilmans:
Do we want to support per-link attributes, or should this helper only be used for "links or routers sharing the same properties"? i.e., do we want to allow:
```python
self.addLinks((r1, r2, {'ip': ("2042:12::1/64", "10.12.0.1/24")}),
              (r1, r3, {'ip': ("2042:12::2/64", "10.12.0.2/24")}), 
              (r2, r3), igp_metric=5)
self.addRouters("r1", ("r2", {'lo_addresses': ["2042:1::1/64", "10.1.1.1/24"]}), config=BorderRouterConfig)
```

__Pros__
  - Explicit support for 'defaults' expressed as kwargs (e.g., the igp_metric) and per-entity settings (e.g., IP addresses) expressed as dict in the entity definitions, instead of spreading/hidding those across multiple callsites
  - Very concise and flexible API

__Cons__
 - This might make the API a bit more contrived/easier to misuse hence require some doc work
 - Errors (i.e., wrong parameter names) will be harder to pinpoint/track down as the stack trace will be opaque

Note the above example can be achieved with the current version of addLinks/addRouters + a succession of addParams calls (i.e., more python boilerplate)

Implementing this would require to match on the types and length of the links/routers parameters, e.g.,
```python
def addRouters(self, *routers, **opts) -> List['RouterDescription']:
        return [self.addRouter(n, {k: v for k, v in itertools.chain(opts.items(), opt.items())})
                for name in routers for n, opt in (name if is_container(name) else (name, {})]

def addLinks(self, *links, **opts) -> List['LinkDescription']:
        return [self.addLink(u, v, {k: v for k, v in itertools.chain(opts.items(), (opt if opt else {}).items())})
                for u, v, *opt in links]
```

jadinm:
We can allow that

> Errors (i.e., wrong parameter names) will be harder to pinpoint/track down as the stack trace will be opaque

Maybe, we can use logging to tackle the last point, i.e., logging the link or router when catching and raising the exception again (see the last commit)

oliviertilmans:
This seems good like this.
We should probably also use this in other example topos.

jadinm:
Jenkins, run tests.

jadinm:
Jenkins, rerun tests.

PR #80, opened 2020-07-07
IPTables: add chain abstraction

oliviertilmans:
Add QoL chain helper to more easily express rules.

I do not have a proper ipmininet install at disposal at the moment, I've likely botched this...

### DIFF
--- a/docs/daemons.rst
+++ b/docs/daemons.rst
@@ -147,21 +147,112 @@ IPTables
 --------
 
 This is currently mainly a proxy class to generate a list of static rules to pass to iptables.
-As such, see `man iptables` and `man iptables-extensions`
-to see the various table names, commands, pre-existing chains, ...
 
 It takes one parameter:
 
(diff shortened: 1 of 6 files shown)


oliviertilmans:
I wanted to stress that NOT was a DSL element and not a python object. Arguably Deny/Allow/transitfiltrr/... Also should be in uppercase (i.e. all terminal constructs).

In the case of NOT this seemed particularly necessary to be as different as possible from the python not built-in...which unfortunately cannot be overloaded through magic methods.

jadinm:
Ok, let's keep "NOT" as it is
I think that's a good idea all things considered

jadinm:
Jenkins, run tests.

jadinm:
Jenkins, rerun tests.

jadinm:
Jenkins, rerun tests.

jadinm:
Jenkins, rerun tests.

PR #85, opened 2020-07-30
Zone delegation and dns root server hints

### LINKED ISSUE #79: The DNS abstraction should support zone delegation
It would be extremely useful to support zone delegation with the DNS abstraction currently provided.

I might be mistaken, but it does not seem possible at the moment.

jadinm:
Fixes #79 

There are two main changes:

1. This makes DNS servers domain automatically delegate subdomains
authority to the appropriate zone if it exists. This behavior can be
disabled and custom zones can be delegated as well.

2. This adds to DNS servers a hint zone for the root domain if defined.
This hint zone has records for the DNS root servers. With these hints,
a client can have access to the root of the DNS from any point in the
hierarchy. This behavior can be disabled.

### DIFF
--- a/docs/daemons.rst
+++ b/docs/daemons.rst
@@ -515,6 +515,16 @@ if they fit in its domain name. Otherwise, another zone will be created.
 
             super().build(*args, **kwargs)
 
+By default, subdomains authority is delegated to the direct child zone name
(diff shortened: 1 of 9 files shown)


jadinm:
Jenkins, run tests.

jadinm:
Jenkins, rerun tests.

PR #90, opened 2020-09-08
Link failure simulation

stoza:


ipmininet/cli.py
```diff
@@ -81,6 +81,25 @@ def do_ping6pair(self, _line):
...
 
+    def do_linkFailure(self,line):
+        """Down the link between 2 nodes"""
+        nodes = line.split(' ')
+        if(len(nodes) != 2):
+            lg.output("Error: specify 2 nodes")
+        else:
+            self.mn.runFailurePlan([(nodes[0],nodes[1])])
+
+    def do_restoreLink(self,line):
```
oliviertilmans, line 92: This limitation should not be there.

I'd suggest to keep track of the failed links in either the Net or CLI object (e.g., a dict[(src,dst)|src|dst]->Link), such that you could do:
linkFailure u v, v w, w x, y z
then
linkRestore v, w, y z
to restore all failed links from v/w and the y z link

ipmininet/cli.py
```diff
@@ -81,6 +81,25 @@ def do_ping6pair(self, _line):
         """Ping (IPv6-only) between first two hosts, useful for testing."""
         self.mn.ping6Pair()
 
+    def do_linkFailure(self,line):
+        """Down the link between 2 nodes"""
+        nodes = line.split(' ')
+        if(len(nodes) != 2):
+            lg.output("Error: specify 2 nodes")
+        else:
+            self.mn.runFailurePlan([(nodes[0],nodes[1])])
```
oliviertilmans, line 90: You defined an API enabling to fail multiple links, you should enable to interface directly with here; e.g.
linkFailure u v , w x
to fail two links.

This would require you to parse the line:
- first looking for link separators
- then for the link descriptions

This is particularly important as in this case this command is strictly equivalent to the already existing:
link u v down

ipmininet/cli.py
```diff
@@ -81,6 +81,25 @@ def do_ping6pair(self, _line):
         """Ping (IPv6-only) between first two hosts, useful for testing."""
         self.mn.ping6Pair()
 
+    def do_linkFailure(self,line):
```
oliviertilmans, line 84: You should document more precisely this command, given you are not only failing the link per say (i.e., simulating a faulty amplifier/switch on the path) but really downing an interface.

This causes internal events to be generated, and applications can listen to them (e.g., with netlink RTM_DELLINK https://linux.die.net/man/7/rtnetlink) causing instant failure recovery for them--i.e., experimenters need to add a L2 switch in between and fail it instead.

ipmininet/ipnet.py
```diff
@@ -549,6 +549,47 @@ def ping6Pair(self):
...
+    def runFailurePlan(self,failure_plan):
+        """this function run a failure plan
+            the plan have to be a list of tuple with the name of routeur between
+            which the link have to be down
+            example of plan:
+            [("R1","R2"),("R8","R9"),("R3","R1")]
+        """
+        print("** Starting failure plan")
+        for link in failure_plan:
+            node1 = self.get(link[0])
```
oliviertilmans, line 562: This can throw an exception if the node name does not exists, which at the moment abruptly terminates everything.
You need to handle this locally here, and likely perform another round of validations in the CLI handler (so your failure plan is a transaction, i.e., all failures are applied or none are).

ipmininet/ipnet.py
```diff
@@ -549,6 +549,47 @@ def ping6Pair(self):
...
+            interfaces = node1.connectionsTo(node2)
+            interfaceNode1 = interfaces[0][0]
+            interfaceNode2 = interfaces[0][1]
+            commande = "ip link set dev " + str(interfaceNode1) + " down"
+            node1.cmd(commande)
+            print("** link between " + link[0] + " and "+ link[1] + " down")
+
+    def restoreLink(self,failure_plan):
+        """function which restore the link
+            does not put a ip address again on
```
oliviertilmans, line 573: That's not very practical, as this then prevents from doing multiple down-up events in a row to, e.g., observe BGP's link flapping protection.

You should store the relevant information about the interface prior to downing it and then restore it.

ipmininet/ipnet.py
```diff
@@ -549,6 +549,47 @@ def ping6Pair(self):
...
+            interfaceNode2 = interfaces[0][1]
+            commande = "ip link set dev " + str(interfaceNode1) + " up"
+            node1.cmd(commande)
+            print("** link between " + link[0] + " and "+ link[1] + " restored")
+
+    """
+    this function run a random failure
+    """
+    def runRandomFailure(self):
+        pass
```
oliviertilmans, line 591: That's quite a big placeholder...

A single random failure can be done with:
random.choice(self.links)

You could however make this more interesting with:
- an optional number of failures (i.e., loop the above)
- a restricted pool of links or nodes whose links can fail (e.g., only AS boundaries).

oliviertilmans (changes requested):
Thanks for the PR.

I would at least try to re-use where possible the Mininet APIs; see
https://github.com/mininet/mininet/blob/master/mininet/cli.py#L268
https://github.com/mininet/mininet/blob/master/mininet/net.py#L892

Arguably, this feature could be implemented as a soft-wrapper around 'link x y down' instead of creating its own set of *Net APIs.

An useful feature that could be added on top of this would be an user-provided timeout for the failure duration, to simulate link flapping.

Please also add at least an example file showing how to use the API, and preferably also a test case (simplest one being a 2-nodes networks where you can assert the non-reachability of each other after issuing the do_failure(...) call, and re-establishment after the cancellation.

ipmininet/ipnet.py
```diff
@@ -549,6 +550,68 @@ def ping6Pair(self):
...
+        number_of_links = len(self.links)
+        if(n > number_of_links):
+            log.output("More link down requested than number of link in the network\n")
+        else:
+            routers = self.routers
+            interfaces_down = {}
+            i = 0
+            while(i < n):
+                router = random.choice(routers)
+                interface = random.choice(router.intfList()[1:]) #remove the loopback interface
```
oliviertilmans, line 605: See random.sample() for a subslice of n random items.

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
...
+                node1 = self.get(link[0])
+                node2 = self.get(link[1])
+                interfaces = node1.connectionsTo(node2)
+                interfaceNode1 = interfaces[0][0]
+                interfaceNode2 = interfaces[0][1]
+                commande = "ip link set dev " + str(interfaceNode1) + " down"
+                node1.cmd(commande)
+                interfaces_down.append(interfaceNode1)
+                print("** link between " + link[0] + " and "+ link[1] + " down")
+            except Exception as e:
```
jadinm, line 575: You could be more specific, it is a KeyError

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
            return: ploss packet loss percentage"""
         return self.pingPair(use_v4=False)
 
+   
+    def runFailurePlan(self,failure_plan):
```
jadinm, line 554: Please use type hinting (https://docs.python.org/3/library/typing.html) for your methods

For instance, here, I think that it is
```suggestion
    def runFailurePlan(self, failure_plan: List[Tuple[str, str]]) -> List[IPIntf]:
```

ipmininet/cli.py
```diff
@@ -81,6 +81,18 @@ def do_ping6pair(self, _line):
         """Ping (IPv6-only) between first two hosts, useful for testing."""
         self.mn.ping6Pair()
 
+    def do_link(self,line):
```
jadinm, line 84: Please use type hinting (https://docs.python.org/3/library/typing.html) for your methods

For instance, here, I think that it is
```suggestion
    def do_link(self, line: str):
```

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
...
+        """this function run a failure plan
+            the plan have to be a list of tuple with the name of routeur between
+            which the link have to be down
+            example of plan:
+            [("R1","R2"),("R8","R9"),("R3","R1")]
+            This will return a list of IPIntf
```
jadinm, line 560: Please use "return: " and specify that the interfaces that are returned are the stopped interfaces

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
            return: ploss packet loss percentage"""
         return self.pingPair(use_v4=False)
 
+   
+    def runFailurePlan(self,failure_plan):
+        """this function run a failure plan
+            the plan have to be a list of tuple with the name of routeur between
+            which the link have to be down
+            example of plan:
+            [("R1","R2"),("R8","R9"),("R3","R1")]
```
jadinm, line 559: Please use "param XXX: description" to specify a parameter

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
...
+                interfaces_down.append(interfaceNode1)
+                print("** link between " + link[0] + " and "+ link[1] + " down")
+            except Exception as e:
+                log.output("ERROR: node "+ str(e) +"\n")
+                return interfaces_down
+        return interfaces_down
+
+    def restoreLink(self,interfaces):
+        """function which restore the link
+           the argument is a list of IPInft object
```
jadinm, line 582: Same for doc and typing (see above)

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
...
+        for interface in interfaces:
+            commande = "ip link set dev " + str(interface) + " up"
+            interface.node.cmd(commande)
+            print("** interfaces " + str(interface) + " up")
+
+    def RandomFailure(self,n):
+        """
+        this function down randomly n link
+        It will return a List of IPintf
+        """
```
jadinm, line 594: Same for doc and typing (see above)

ipmininet/cli.py
```diff
@@ -81,6 +81,18 @@ def do_ping6pair(self, _line):
...
+        """ down/up the link between 2 specified routers, can specify multiple
+            multiple router at once by doing link u v, w x [down/up]
```
jadinm, line 86: Please use "param XXX: description" to specify a parameter

ipmininet/ipnet.py
jadinm: Moreover this does not solve the comment of @oliviertilmans 

> This can throw an exception if the node name does not exists, which at the moment abruptly terminates everything.
You need to handle this locally here, and likely perform another round of validations in the CLI handler (so your failure plan is a transaction, i.e., all failures are applied or none are).

If an exception is triggered here, you will still have links that are shutdown and links that are not.
You need to first check that interfaces and links exist before executing the commands to remove them

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
...
+        """function which restore the link
+           the argument is a list of IPInft object 
+        """
+        print("** starting restoring link")
+        for interface in interfaces:
+            commande = "ip link set dev " + str(interface) + " up"
+            interface.node.cmd(commande)
+            print("** interfaces " + str(interface) + " up")
+
+    def RandomFailure(self,n):
```
jadinm, line 590: You could add an optional parameter that restrict the links that can fail to match https://github.com/cnp3/ipmininet/pull/90#discussion_r492015364

So if the parameter is defined, you use it instead of self.links and otherwise, you stay with self.links :)

jadinm (changes requested):
Thanks for the PR :)
I have some comments as well (see inline)

It would be nice to document your modifications to the CLI as well as the failure plan in the documentation (in the docs folder)

And you can add some tests as well in folder ipmininet/tests and an example in ipmininet/example as suggested y @oliviertilmans:
> Please also add at least an example file showing how to use the API, and preferably also a test case (simplest one being a 2-nodes networks where you can assert the non-reachability of each other after issuing the do_failure(...) call, and re-establishment after the cancellation.

On a minor note, please use standard spacing conventions for parameters in Python: one space after each comma

stoza:
The issue of adding a examples is that all the stuff of downing link happen after the net.start() call.
And the way in which the examples are designed does not allow to instantiate myself the IPnet, unless modifying the __main__.py file which is not practical and not useful for showing how the things work

ipmininet/ipnet.py
```diff
@@ -549,6 +550,79 @@ def ping6Pair(self):
...
+
+    def RandomFailure(self, n, weak_links=None) -> List[IPIntf]:
+        """ this function down randomly n link
+            :param n: the number of link to be downed
+            :param weak_links: an optional parameter which specify a list of IPIntf in
+                                which the n links will be downed
+            :return: a list of IPIntf which have been downed
+        """
+        if weak_links is None:
+            all_links = self.links
```
jadinm, line 598: I was actually thinking to something much simpler than what you implemented:

```suggestion
            all_links = weak_links if weak_links is not None else self.links
```

The idea is just to restrict the links on which random runs, not to shutdown all the links in a method called "RandomFailure". It would be strange

jadinm (commented):
> The issue of adding a examples is that all the stuff of downing link happen after the net.start() call.
> And the way in which the examples are designed does not allow to instantiate myself the IPnet, unless modifying the ``__main__.py`` file which is not practical and not useful for showing how the things work

Actually, you can use ``post_buid(self, net: IPNet)`` method of the example topology for that. The network, nodes and links are created. The daemons are not launched though.
You could couple that to a timer before executing your failures.

For the CLI commands and parameters, they can be documented in the main documentation and in the ``ipmininet/examples/README.md`` file. In the latter one, you can give an example of link to kill.

ipmininet/ipnet.py
stoza: I'm not sure to understand the issue here. not all the link are downed here, either no restricted link are specified (the weak_links which is a list of IPIntf) and then only n links of all the links are downed using self.links(which is a List of IPLink). Or a subset of links to down is specify using the weak_links parameter and then only n links are downed in this list.
The issue why this implementation seems so complicated is because self.links return a list of IPLinks while weak_links are a list of IPIntf. Because I've not found in the doc how we can retrieve a list of IPLink based on a node.

ipmininet/ipnet.py
```diff
@@ -549,6 +550,79 @@ def ping6Pair(self):
...
+        """function which restore the link
+            :param interfaces: a List of IPIntf
+        """
+        log.output("** starting restoring link\n")
+        for interface in interfaces:
+            commande = "ip link set dev " + str(interface) + " up"
+            interface.node.cmd(commande)
+            log.output("** interfaces " + str(interface) + " up\n")
+
+    def RandomFailure(self, n: int, weak_links:List[IPIntf]=None) -> List[IPIntf]:
```
jadinm, line 590: You need to specify ``Optional`` if a parameter can be ``None``

```suggestion
    def RandomFailure(self, n: int, weak_links:Optional[List[IPIntf]]=None) -> List[IPIntf]:
```

ipmininet/ipnet.py
jadinm: > Because I've not found in the doc how we can retrieve a list of IPLink based on a node.

You can use ``intf.link`` to get the parent link of an interface

> The issue why this implementation seems so complicated is because self.links return a list of IPLinks while weak_links are a list of IPIntf.

To be consistent, ``weak_links`` should be a list of ``IPLink`` then
And this should simplify your implementation

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
...
+        for link in failure_plan:
+            try:
+                node1 = self.get(link[0])
+                node2 = self.get(link[1])
+                interfaces = node1.connectionsTo(node2)
+                interfaceNode1 = interfaces[0][0]
+                interfaceNode2 = interfaces[0][1]
+                interfaces_down.append(interfaceNode1)
+            except KeyError as e:
+                log.output("ERROR: node "+ str(e) +"\n")
```
jadinm, line 572:
```suggestion
                log.error("ERROR: node "+ str(e) +"\n")
```

ipmininet/router/__router.py
```diff
@@ -193,7 +193,7 @@ def __init__(self, name,
                  lo_addresses: Sequence[Union[str, IPv4Interface,
                                               IPv6Interface]] = (),
                  *args, **kwargs):
-        """:param password: The password for the routing daemons vtysh access
+        """:param password: The password for the routing daemons vtysh access
```
jadinm, line 196:
```suggestion
        """:param password: The password for the routing daemons vtysh access
```

ipmininet/cli.py
```diff
@@ -81,6 +81,18 @@ def do_ping6pair(self, _line):
         """Ping (IPv6-only) between first two hosts, useful for testing."""
         self.mn.ping6Pair()
 
+    def do_link(self, line):
```
jadinm, line 84:
```suggestion
    def do_link(self, line: str):
```

ipmininet/ipnet.py
```diff
@@ -549,6 +550,64 @@ def ping6Pair(self):
...
+        """ this function down randomly n link
+            :param n: the number of link to be downed
+            :param weak_links: an optional parameter which specify a list of IPIntf in
+                                which the n links will be downed
+            :return: a list of IPIntf which have been downed
+        """
+        all_links = weak_links if weak_links is not None else self.links
+        number_of_links = len(all_links)
+        if(n > number_of_links):
+            log.output("More link down requested than number of link in the network\n")
```
jadinm, line 600:
```suggestion
            log.warning("More link down requested than number of link in the network\n")
```

ipmininet/cli.py
```diff
@@ -81,6 +81,18 @@ def do_ping6pair(self, _line):
...
+        """ down/up the link between 2 specified routers, can specify multiple
+            multiple link
```
jadinm, line 86: Could you describe the format of line here ?

jadinm (commented):
A few more comments on your last changes

ipmininet/cli.py
stoza: Yes i didn't do it because it's not something which is done in the other things

jadinm:

jadinm (approved):

jadinm:
Jenkins, run tests.

jadinm:
Jenkins, run tests.

jadinm:
Jenkins, run tests.

jadinm:
Jenkins, run tests.

jadinm:
Jenkins, run tests.

PR #91, opened 2020-09-16
[build_vm] Fix checkout of ipmininet branch

butjar:
The build script fails when checking out an existing branch with
following message: 'fatal: A branch named 'master' already exists.'

Use git's -B option that works also when the branch already exists.

### DIFF
--- a/ipmininet/install/build_vm.sh
+++ b/ipmininet/install/build_vm.sh
@@ -37,7 +37,11 @@ sudo apt remove -yq python3-pip
 
 [ ! -d "$IPMN_DIR" ] && git clone ${IPMN_REPO} ${IPMN_DIR}
 pushd ${IPMN_DIR}
-git checkout -B ${IPMN_BRANCH} -t origin/${IPMN_BRANCH}
+git checkout -B ${IPMN_BRANCH} -t origin/${IPMN_BRANCH} || >&2 cat <<-EOF
+	WARN: Command 'git checkout -B ${IPMN_BRANCH} -t origin/${IPMN_BRANCH}' failed.
+	Please check the ipmininet repository.
+EOF
+
 sudo pip3 install .
 sudo python3 -m ipmininet.install -af6
 popd
--- a/ipmininet/install/build_vm.sh
+++ b/ipmininet/install/build_vm.sh
@@ -37,7 +37,7 @@ sudo apt remove -yq python3-pip
 
 [ ! -d "$IPMN_DIR" ] && git clone ${IPMN_REPO} ${IPMN_DIR}
 pushd ${IPMN_DIR}
-git checkout -t origin/${IPMN_BRANCH}
+git checkout -B ${IPMN_BRANCH} -t origin/${IPMN_BRANCH}
 sudo pip3 install .
 sudo python3 -m ipmininet.install -af6
 popd


butjar:
Log after change: 
```
...
    terranet-base: Processing triggers for man-db (2.8.3-2ubuntu0.1) ...
    terranet-base: Cloning into 'ipmininet'...
    terranet-base: ~/ipmininet ~ /usr/local/src
    terranet-base: Reset branch 'master'
    terranet-base: Branch 'master' set up to track remote branch 'master' from 'origin'.
    terranet-base: Your branch is up to date with 'origin/master'.
    terranet-base: Processing /root/ipmininet
...
```

butjar:
I could swear this worked before. Maybe a new git version has a more restrictive checkout behavior.

ipmininet/install/build_vm.sh
```diff
@@ -37,7 +37,7 @@ sudo apt remove -yq python3-pip
 
 [ ! -d "$IPMN_DIR" ] && git clone ${IPMN_REPO} ${IPMN_DIR}
```
oliviertilmans, line 38:
```suggestion
[ ! -d "$IPMN_DIR" ] && git clone ${IPMN_REPO} ${IPMN_DIR} --branch ${IPMN_BRANCH}
```

ipmininet/install/build_vm.sh
```diff
@@ -37,7 +37,7 @@ sudo apt remove -yq python3-pip
...
 pushd ${IPMN_DIR}
-git checkout -t origin/${IPMN_BRANCH}
+git checkout -B ${IPMN_BRANCH} -t origin/${IPMN_BRANCH}
```
oliviertilmans, line 40:
```suggestion
```

oliviertilmans (changes requested):
While your change is correct, assuming that newer git versions have become stricter, I'd rather instead immediately clone/checkout the repo correctly thank cloning then switching branches around in two iterations.

ipmininet/install/build_vm.sh
butjar: This assumes that we always execute the script in a fresh environment. The script was written so it might handle multiple invocations (not sure if this really works with `sudo python3 -m ipmininet.install -af6`). I can implement your suggestion if the script should not be idempotent. 

Another alternative is to trash an existing directory:
```bash
rm -rf ${IPMN_DIR}
git clone ${IPMN_REPO} ${IPMN_DIR} -b ${IPMN_BRANCH}
```

oliviertilmans: That's a very good point indeed.

Should an existing repo exist, the current checkout can however fail if there are uncommitted changes that would be lost. Your solution with an added
```bash
|| echo "! Could not reset the ipmininet git repository!"
```
Should be sufficient then, i.e., have the existing behavior and warn if something is not right.

butjar: I tested the behavior:
```
root@mininet-vm:~/ipmininet# git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
root@mininet-vm:~/ipmininet# echo FOO >> README.md
root@mininet-vm:~/ipmininet# git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   README.md

no changes added to commit (use "git add" and/or "git commit -a")
root@mininet-vm:~/ipmininet# git checkout -B master -t origin/master
M       README.md
Branch 'master' set up to track remote branch 'master' from 'origin'.
Reset branch 'master'
Your branch is up to date with 'origin/master'.
root@mininet-vm:~/ipmininet# git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   README.md

no changes added to commit (use "git add" and/or "git commit -a")
root@mininet-vm:~/ipmininet#

```
Unstanged changes are kept and the checkout works fine. New commits will be reset. However, I will add a general warning message if the checkout fails (helps debugging when we experience different git behavior again :sweat_smile:).

ipmininet/install/build_vm.sh
butjar: @oliviertilmans Please review the commit I added 6d84c250aebd161717315b86aa27dc256b79cb9c.

ipmininet/install/build_vm.sh
oliviertilmans: That's indeed the case for tracked files common to all branches, but not if the changes are not tracked in the starting (recent examples in point: having someone accidentally commit files generated by the build system in its feature branch).
This would show up as (gimmic example: I stopped tracking LICENSE.md in my branch and made some changes to it):
```bash
❯ git checkout -B master origin/master 
error: The following untracked working tree files would be overwritten by checkout:
        LICENSE.md
Please move or remove them before you switch branches.
Aborting
```

I agree that is not a common scenario, but at the same time, if someone is building from custom repo/branches, that's exactly the type of person that could run into this (e.g., by trying locally changes from PR).

oliviertilmans:
Thanks @butjar !

ipmininet/install/build_vm.sh
butjar: I was able to reproduce the scenario:
```
root@mininet-vm:~/ipmininet# git checkout -b foo
Switched to a new branch 'foo'
root@mininet-vm:~/ipmininet# git rm --cached LICENSE.md
rm 'LICENSE.md'
root@mininet-vm:~/ipmininet# echo FOO > LICENSE.md
root@mininet-vm:~/ipmininet# git commit -m FOO
[foo 47aa694] FOO
 1 file changed, 339 deletions(-)
 delete mode 100644 LICENSE.md
root@mininet-vm:~/ipmininet# git checkout -B master origin/master
error: The following untracked working tree files would be overwritten by checkout:
        LICENSE.md
Please move or remove them before you switch branches.
Aborting
```

Good to know, thanks for this detail.

PR #94, opened 2020-10-19
Plot

Ananas120:
Plot example code
To test it, i put the topology from the TP in examples/topo_tp.json so you can try the code by running : 
```python
from ipmininet.plot_utils import to_graphviz
g = to_graphviz(‘ipmininet/examples/topo_tp.json’, filename = ‘plot_test.gv’)
g.view()
g # if you are in a jupyter notebook, it will show the plot
```

### DIFF
--- a/ipmininet/examples/topo_tp.json
+++ b/ipmininet/examples/topo_tp.json
@@ -1,0 +1,56 @@
+{
+    "subnets" : {
+        "as1_lan" :{
+            "address" :"cafe:babe:dead:beaf::/64",
+            "hosts" : {"as1_h1" : "as1_s2"}
+        },
+        "as2_lan" :{
+            "address":"c1a4:4ad:c0ff:ee::/64",
+            "hosts" : {"as2_h2" : ["as2_cl1", "as2_cl2"]}
+        }
+    },
+    "AS" : {
+        "AS1" : {
+            "routers" : {
+                "as1_rr1" : {
+                    "clients" : ["as1_s1"],
+                    "daemons" : {"bgp" : {"networks":["as1_lan"]}}
+                },
+                "as1_rr2" : {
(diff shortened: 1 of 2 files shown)


jadinm:
I am sceptical about using a json file as an argument
Wouldn't it be better to enable users to display the graph from an IPTopo object directly than requiring an additional file from them ?

Also, it would be better not to have hardcoded paths such as the variable ```_graphviz_path```

jadinm:

Ananas120:
Yes indeed, my first target was to do it based on a IPTopo class but i didn’t find the information of AS in the class
Furthermore, the « json » idea is also related to my class JSONTopo i created for the project which also take a json file and automatically create the topology (which is far more powerful than hardcode all routers / hosts / AS / ...) but i will post it only at the end of the project 

So the idea of this plot was to use the json file to 1) create the topology based on JSONTopo class and 2) plot the topology based on the same json file and 3) based on the JSONTopo class, get many informations about the topology (number of iBGP / eBGP connections, number of links / RR / ...)

jadinm:
> Yes indeed, my first target was to do it based on a IPTopo class but i didn’t find the information of AS in the class
> Furthermore, the « json » idea is also related to my class JSONTopo i created for the project which also take a json file and automatically create the topology (which is far more powerful than hardcode all routers / hosts / AS / ...) but i will post it only at the end of the project

You can have access to this by using the instance variable ``topology.overlays``
Then, you can check the type of the ```Overlay```
If this is an AS object with ```isinstance()```, then you have access to the AS variable :)

Ananas120:
Ok thank you, then i will try to convert a IPTopo to a json file it can be good to visualize the topology etc !  
I will not have the time to do it before the end of the project but maybe after when posting the JSONTopo

PR #95, opened 2020-10-26
hardcoded address replaced by router-id

hlloreda:
In order to avoid having the same cluster ID for all the different RRs, we have set the cluster-id to the router-id.


jadinm:
Thank you for the PR :+1: 

Jenkins, ok to test.

jadinm:

jadinm:
ok to test

jadinm:
Jenkins, run tests.

PR #98, opened 2020-11-14
Fix CommunityList example in daemons.rst

fdardenne:
Hello,

This PR meant to fix an error in the documentation for the BGP Daemon example.
The CommunityList class has 3 optional parameters, the actual example set the "name" parameter to 'local-pref' and the "action" parameter to "2:80". Instead of setting the "name" and the "community" parameters.

Resulting in this line in the router configuration:
`bgp community-list loc-pref 2:80 1:0 `
However it should be:
`bgp community-list loc-pref PERMIT 2:80`

### DIFF
--- a/docs/daemons.rst
+++ b/docs/daemons.rst
@@ -104,7 +104,7 @@ The following code shows how to use all these abstractions:
             all_al = AccessList('all', ('any',))
 
             # Add a community list to as2r1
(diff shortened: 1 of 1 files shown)


jadinm:

jadinm:
Thank you for the PR :+1:

PR #110, opened 2021-05-25
Add support for FRR vtysh

gromit1811:
vtysh in FRR by default uses Unix domain sockets in /var/run/frr to access
the CLI of routing daemons. So far, running vtysh always caused it to
connect to the most recently started FRR daemons because they created their
sockets last and thus overwrote those of earlier daemons. This commit makes
/var/run/frr a private directory specific to each router, so running vtysh
in the router's namespace makes it connect to the daemons responsible for
that router.

I don't know which directory was used by the pre-FRR Quagga (maybe
/var/run/quagga?), but it proabably simply could be added to the list and
we'd have vtysh support both for FRR and Quagga in case somebody is still
using the latter.

Signed-off-by: Martin Buck <mb-tmp-tvguho.pbz@gromit.dyndns.org>


jadinm:
Jenkins, run tests

